PR #9, opened 2025-10-24
Fix macOS Seatbelt read bypass via file movement

ddworken:
## Summary

Fixes a security vulnerability where files protected by read deny rules could be exfiltrated by moving them to readable locations using the `mv` command.

## Vulnerability Details

The `denyOnly` read restriction configuration only generated `(deny file-read*)` rules in Seatbelt profiles, which did not block the `rename()` syscall used by `mv` operations.

### Attack Vectors

**Method 1: Move the file directly**
```bash
$ mv ~/.aws/credentials ~/project/stolen.txt
$ cat ~/project/stolen.txt  # Contents now readable
```

**Method 2: Move an ancestor directory**
```bash
$ mv ~/.aws ~/moved-aws
$ cat ~/moved-aws/credentials  # Contents now readable
```

## Root Cause

- Seatbelt `(deny file-read*)` rules only block read operations, not rename operations
- The `rename()` syscall is considered a write operation on parent directories
- When ancestor directories are moved, the deny rule continues to reference the old absolute path

## Fix

Added `(deny file-write-unlink)` rules for read-denied paths to block rename/move operations:
1. Block moving/renaming the denied files/directories themselves
2. Block moving all ancestor directories (up to but not including root)

### Implementation

- Added `getAncestorDirectories()` helper function (src/sandbox/macos-sandbox-utils.ts:117-137)
- Modified `generateReadRules()` to add file-write-unlink deny rules (src/sandbox/macos-sandbox-utils.ts:142-220)
- Works with both literal paths and glob patterns

### Example

For a denied path `/private/tmp/test-denied/secret.txt`:
- Deny file-write-unlink on `/private/tmp/test-denied/secret.txt` (the file itself)
- Deny file-write-unlink on `/private/tmp/test-denied` (parent directory)
- Deny file-write-unlink on `/private/tmp` (grandparent directory)

## Test Coverage

Added comprehensive test suite in `test/sandbox/macos-seatbelt.test.ts`:
- Verifies direct file moves are blocked
- Verifies parent directory moves are blocked
- Verifies glob patterns work correctly
- Verifies read restrictions still function

All tests pass on macOS.

## Platform

- **Affected**: macOS (using Seatbelt/sandbox-exec)
- **Not affected**: Linux (uses different sandboxing mechanism)


src/sandbox/macos-sandbox-utils.ts
```diff
@@ -125,13 +148,62 @@ function generateReadRules(
...
+      // Block moving/renaming files matching this pattern
+      rules.push(
+        `(deny file-write-unlink`,
+        `  (regex ${escapePath(regexPattern)})`,
+        `  (with message "${logTag}"))`,
+      )
+
+      // For glob patterns, extract the static prefix and block ancestor moves
+      // Remove glob characters to get the directory prefix
+      const staticPrefix = normalizedPath.split(/[*?\[]/)[ 0]
```
ollie-anthropic: is this missing the closing bracket in the regex?  like:

```const staticPrefix = normalizedPath.split(/[*?\[\]]/)[ 0]```

**Review by ollie-anthropic — COMMENTED** [review]
Thanks for working on this ! Gave this a little test and realized it will also apply for writes ! so we should add that in - i think we may also be able to factorize some of the duplicated logic

e.g. you can block writes to a/b, but then do mv a c, then write to c/b, then mv c a, and it allows you to write to a/b :woe:

File: src/sandbox/sandbox-utils.ts [hunk 2]
```diff
@@ -87,8 +87,27 @@ export function normalizePathForSandbox(pathPattern: string): string {
     normalizedPath = path.resolve(cwd, pathPattern)
   }
 
-  // For glob patterns, don't try to resolve symlinks (they don't exist as literal paths)
+  // For glob patterns, resolve symlinks for the directory portion only
   if (containsGlobChars(normalizedPath)) {
+    // Extract the static directory prefix before glob characters
+    const staticPrefix = normalizedPath.split(/[*?\[]/)[ 0]
```
ollie-anthropic: @ddworken looks like the other comment was resolved, but see the same pattern - are we missing the \[ ?

ddworken: Fixed!

ollie-anthropic (approved):
tyty

PR #16, opened 2025-10-31
Add binShell parameter to allow specifying shell for sandboxed commands

ddworken:
## Summary
- Adds optional `binShell` parameter to `LinuxSandboxParams` and `MacOSSandboxParams`
- Allows users to specify which shell (zsh, bash, etc.) to use when executing commands in the sandbox
- Defaults to 'bash' for backward compatibility
- Enables shell-specific features like aliases and snapshots to work properly

## Changes
- Updated `src/sandbox/linux-sandbox-utils.ts` to accept and use `binShell` parameter
- Updated `src/sandbox/macos-sandbox-utils.ts` to accept and use `binShell` parameter  
- Updated `src/sandbox/sandbox-manager.ts` to thread through the `binShell` parameter


ollie-anthropic (approved):
nice - ty

PR #22, opened 2025-11-04
Add configurable httpProxyPort and socksProxyPort to network schema

ddworken:
## Summary
- Added optional `httpProxyPort` and `socksProxyPort` fields to `NetworkConfigSchema`
- When specified, the library uses the external proxy port instead of starting its own proxy
- Each proxy (HTTP and SOCKS) can independently be external or local
- External proxies are responsible for handling domain filtering (allowedDomains/deniedDomains are ignored for external proxies)

## Changes
**src/sandbox/sandbox-config.ts:**
- Added `httpProxyPort` and `socksProxyPort` fields with validation (1-65535)

**src/sandbox/sandbox-manager.ts:**
- Updated `HostNetworkManagerContext` to track whether each proxy is local or external
- Modified `initialize()` to conditionally start proxies based on config
- Updated `reset()` to only close locally-started proxies

**test/configurable-proxy-ports.test.ts:**
- Added 9 integration tests covering various proxy configurations
- Includes end-to-end test with actual external proxy server that verifies requests route through external proxy

## Test Results
✅ All 78 tests pass (9 new tests + 69 existing tests)
✅ No regressions introduced


src/sandbox/sandbox-manager.ts
```diff
@@ -219,11 +221,32 @@ async function initialize(
...
-        startHttpProxyServer(sandboxAskCallback),
-        startSocksProxyServer(sandboxAskCallback),
-      ])
+      // Conditionally start proxy servers based on config
+      let httpProxyPort: number
+      let httpProxyIsLocal: boolean
+      if (config.network.httpProxyPort !== undefined) {
+        // Use external HTTP proxy
+        httpProxyPort = config.network.httpProxyPort
+        httpProxyIsLocal = false
```
ollie-anthropic: tiny nit: i got a little confused as proxies will still be local in both cases (as you're only specifying the port, it's always going to be localhost)

could we name this something like: isManagedByCC? but do we need this value? can we not have an undefined httpProxyServer instead 👀

ollie-anthropic (approved):
super nit but else lgtm!!

PR #32, opened 2025-11-10
Pin Bun 1.3.1 to fix CI on Linux

kmckiern:
Pin the CI workflow to Bun `1.3.1` instead of `latest`.

Bun 1.3.2 now auto-cleans dangling processes in `bun test` ([see](https://github.com/oven-sh/bun/blob/bun-v1.3.2/src/bun.js/test/Execution.zig#L180)). This kills the long-lived socat processes keeping `/tmp/claude-*.sock` alive.

Pinning is a short term fix. The longer term fix probably involves keeping the socat processes (and their sockets) alive.


ollie-anthropic (approved):
thank you!! i can follow up with that recommendation

PR #48, opened 2025-11-24
Fix empty allowedDomains to block network as documented

ddworken:
## Summary

- Fixed logic where `allowedDomains: []` was incorrectly allowing full network access instead of blocking all network
- Documentation states "Empty array = no network access" - code now matches this behavior
- Separated `needsNetworkRestriction` from `needsNetworkProxy` to properly handle the "block all" case

## Test plan

- [x] Added unit tests for platform-specific functions with empty domains
- [x] Added integration tests verifying network is blocked when `allowedDomains: []`
- [x] All existing tests pass (116 tests)
- [x] TypeScript type checking passes


ollie-anthropic (approved):
thank you - and for adding e2e tests - crazy that this slipped through

PR #57, opened 2025-11-30
Adding option to allow pseudo-terminal for sandboxing (macOS)

hackyon-anthropic:


ddworken (approved):
Would you mind also adding a test for this?

PR #80, opened 2026-01-06
Harden sandbox deny path handling for non-existent files

ddworken:
## Summary

Adds defense-in-depth for sandbox deny path handling. Previously, non-existent paths in the deny list were skipped since bwrap cannot ro-bind a file that doesn't exist. This change mounts `/dev/null` at the first non-existent path component, preventing the path from being created.

## Changes

- Add `findFirstNonExistentComponent` helper to find the appropriate mount point
- Mount `/dev/null` read-only at the first missing component to block path creation
- Add tests for non-existent deny path protection

## Test plan

- [x] Build passes
- [x] Unit tests pass
- [ ] Linux integration tests (CI)

🤖 Generated with [Claude Code](https://claude.ai/code)


ollie-anthropic (approved):
nice thank you!

PR #84, opened 2026-01-08
Reuse `process.env.TMPDIR` if set

JacquesAnthropic:
Currently we're hardcoding `/tmp/claude`, this reuses `TMPDIR` if set before falling back to `/tmp/claude`


felixrieseberg (approved):

ollie-anthropic (approved):
approved - just need to fix the merge conflicts >.<

PR #90, opened 2026-01-13
add fd 3 control channel for dynamic config updates

loc:
Motivating this change is the desire to have a permission prompt in Cowork that we send to the user that asks: "Do you want to allow Claude to access example.com?" To do this we need to dynamically define network rules safely (without risk of privilege elevation)

Left unsolved in this PR is how to communicate that though you can theoretically update the entire config, some of the configuration options are set at boot time and thus changing them will do nothing. We could scope the updates to just the network config, but I thought I'd get your input before doing that.

## Summary

Replace file-based config watching with a secure fd 3 control channel for dynamic sandbox configuration updates.

## Problem

The initial file watcher approach (`~/.srt-settings.json`) had a privilege escalation risk:
- Any process that can write to the config file could relax sandbox restrictions
- A compromised process or sandbox escape could modify the file to gain network access

## Solution

Use fd 3 (file descriptor 3) as a control channel for config updates. This is secure because:
- Only the parent process can write to the child's fd 3 (pipe is unidirectional)
- A sandboxed process cannot write to srt's fd 3 from outside
- Clean privilege separation tied to process hierarchy
- stdin remains available for the sandboxed command (interactive use still works)

## Protocol

JSON Lines on fd 3:
- Each line is a complete `SandboxRuntimeConfig` JSON object
- Full replacement semantics (not patches)
- Invalid JSON/schema is logged and ignored (keeps old config)
- Empty lines are ignored

## Usage

```bash
# Parent process spawns srt with fd 3 as a writable pipe
srt --control-fd 3 -- my-command

# Parent sends config updates via the pipe
echo '{"network":{"allowedDomains":["example.com"],"deniedDomains":[]},"filesystem":{"denyRead":[],"allowWrite":[],"denyWrite":[]}}' >&3
```

From Node.js:
```typescript
const srt = spawn('srt', ['--control-fd', '3', '--', 'bash'], {
  stdio: ['inherit', 'inherit', 'inherit', 'pipe']  // fd 3 is a writable pipe
})

// Send config update
srt.stdio[3].write(JSON.stringify(config) + '\n')
```

## Changes

- **`src/cli.ts`**: Add `--control-fd <fd>` option, set up readline listener for JSON lines
- **`src/utils/config-loader.ts`**: Add `loadConfigFromString()`, remove `watchConfigFile()`
- **`test/control-fd.test.ts`**: Integration tests for the control channel
- **`test/cli-config-loading.test.ts`**: Unit tests for `loadConfigFromString()`

## Additional Fix

When `network` config is specified with an empty `allowedDomains: []`, traffic is now always routed through the proxy. This allows `updateConfig()` to enable network access later (previously, empty allowlist meant no proxy setup).

## Testing

- 220 tests pass
- Integration tests verify config updates work via fd 3
- Backward compatibility: without `--control-fd`, srt works exactly as before


src/sandbox/sandbox-manager.ts
```diff
@@ -520,19 +520,15 @@ async function wrapWithSandbox(
...
   const needsNetworkRestriction = hasNetworkConfig
 
-  // Network PROXY is only needed when there are domains to filter
-  // If allowedDomains is empty, we block all network and don't need the proxy
-  const needsNetworkProxy = allowedDomains.length > 0
+  // Network PROXY is needed whenever network config is specified
+  // Even with empty allowedDomains, we route through proxy so that:
+  // 1. updateConfig() can enable network access for already-running processes
+  // 2. The proxy blocks all requests when allowlist is empty
+  const needsNetworkProxy = hasNetworkConfig
```
loc, line 531: this is sorta unrelated to the PR, but i found this pretty confusing when testing

loc: another benefit is that now if you have no allowedDomains, curl will complain (correctly) about it being blocked rather than having no net at all.

ddworken: Makes perfect sense to me, thank you for fixing!

ddworken (approved):

Issue #85, opened 2026-01-09
srt generates dotfiles and directories on command execution


la-j:
## Description
Running any command via `srt` generates numerous dotfiles/directories in the current working directory. Files appear regardless of command executed.

## Files Created
```
.bash_profile
.bashrc
.claude/
.env
.git/
.gitconfig
.gitmodules
.idea/
.mcp.json
.profile
.ripgreprc
.vscode/
.zprofile
.zshrc
config/
```

## Reproduction Steps
1. Create empty directory with only `srt-settings.json`
2. Run `srt --settings srt-settings.json "ls"`
3. Check directory contents with `ls -a`

## Observed Behavior
All dotfiles listed above appear in working directory after command execution.

## Expected Behavior
`srt` should not create dotfiles in working directory. These appear to be sandbox/container home directory files leaking into host working directory.

## Environment
- **OS**: Ubuntu 24.04.3 LTS (Noble Numbat)
- **Kernel**: 6.8.0-90-generic x86_64
- **srt version**: 1.0.0 (installed via npm at ~/.npm_global/bin/srt)

### srt-settings.json
```json
{
  "network": {
    "allowedDomains": [
      "github.com",
      "*.github.com",
      "lfs.github.com",
      "api.github.com",
      "npmjs.org",
      "*.npmjs.org"
    ],
    "deniedDomains": [
      "malicious.com"
    ],
    "allowUnixSockets": [
      "/var/run/docker.sock"
    ],
    "allowLocalBinding": false
  },
  "filesystem": {
    "denyRead": [
      "~/.ssh"
    ],
    "allowWrite": [
      ".",
      "src/",
      "test/",
      "/tmp"
    ],
    "denyWrite": [
      ".env",
      "config/production.json"
    ]
  },
  "ignoreViolations": {
    "*": [
      "/usr/bin",
      "/System"
    ],
    "git push": [
      "/usr/bin/nc"
    ],
    "npm": [
      "/private/tmp"
    ]
  },
  "enableWeakerNestedSandbox": false
}
```

### AppArmor Profile (`/etc/apparmor.d/bwrap`)
```
abi <abi/4.0>,
include <tunables/global>

profile bwrap /usr/bin/bwrap flags=(unconfined) {
  userns,

  # Site-specific additions and overrides. See local/README for details.
  include if exists <local/bwrap>
}
```

la-j:
```bash
[...]

PR #91, opened 2026-01-14
Fix dotfile leak by skipping non-existent deny paths

la-j:
## Reason

Running srt generates empty dotfiles (.bashrc, .gitconfig, etc.) in working directory. Bubblewrap's `--ro-bind /dev/null /path` creates empty files on host as mount points that persist after sandbox exits.

## Description

- Skip non-existent deny paths instead of mounting /dev/null over them
- Remove findFirstNonExistentComponent() function (dead code)
- Remove non-existent deny path protection tests

Fixes #85


ollie-anthropic (approved):
Thank you! Verified the fix works. Tested on Linux:

**Main (broken):** Creates 13 empty dotfiles (.bashrc, .gitconfig, .claude, etc.) in working directory

**This PR:** No dotfiles created, directory stays clean

The fix is simple and correct - just skip non-existent deny paths instead of trying to mount /dev/null over them.

PR #99, opened 2026-01-21
fix: support WSL2 sandboxing, reject WSL1

ollie-anthropic:
## Summary
WSL2+ uses the same bubblewrap sandboxing as Linux, so it should be supported. WSL1 does not support bubblewrap properly.

## Changes
- Add `getWslVersion()` to detect WSL version from `/proc/version`
- Update `isSupportedPlatform()` to check for WSL1 and reject it
- Remove 'wsl' from Platform type (`getPlatform` returns 'linux' for WSL)
- Update comments to clarify Linux/WSL glob pattern handling

## Testing
- All 225 tests pass
- Logic: macOS ✓, non-WSL Linux ✓, WSL2+ ✓, WSL1 ✗


src/utils/platform.ts
```diff
@@ -2,13 +2,51 @@
...
 
+import * as fs from 'fs'
+
 export type Platform = 'macos' | 'linux' | 'windows' | 'unknown'
 
+/**
+ * Get the WSL version (1 or 2+) if running in WSL.
+ * Returns undefined if not running in WSL.
+ */
+export function getWslVersion(): string | undefined {
```
ollie-anthropic, line 13: took this from claude-cli code :)

src/sandbox/sandbox-manager.ts
```diff
@@ -874,7 +878,7 @@ export interface ISandboxManager {
     sandboxAskCallback?: SandboxAskCallback,
     enableLogMonitor?: boolean,
   ): Promise<void>
-  isSupportedPlatform(platform: Platform): boolean
+  isSupportedPlatform(): boolean
```
ollie-anthropic, line 881: api changed here but i really think it's unused and this is cleaner

dylan-conway (approved):
nice

src/sandbox/sandbox-manager.ts
```diff
@@ -314,9 +315,13 @@ async function initialize(
   await initializationPromise
 }
 
-function isSupportedPlatform(platform: Platform): boolean {
-  const supportedPlatforms: Platform[] = ['macos', 'linux']
-  return supportedPlatforms.includes(platform)
+function isSupportedPlatform(): boolean {
+  const platform = getPlatform()
```
ant-kurt, line 319: This could add latency due to containing a sync FS call (reading the /proc/version)

PR #108, opened 2026-01-30
Harden sandbox by removing unnecessary trustd.agent mach-lookup

ddworken:
Removes the `com.apple.trustd.agent` mach-lookup from the sandbox profile. This service runs outside the sandbox and is not needed by sandboxed processes — TLS is handled by the proxy layer, not by the sandboxed process directly.

Tested extensively: curl, wget, git, npm install, pip install, cargo build, python/ruby/node HTTPS, compilation toolchains, keychain operations, and all existing unit + integration tests pass without it.


ollie-anthropic (approved):
thank you!

Issue #118, opened 2026-02-05
Go programs fail TLS verification after trustd.agent removal (breaks httpProxyPort for gh, gcloud, etc.)


patrick-premont:
## Summary

PR #108 removed `com.apple.trustd.agent` from the mach-lookup allowlist, which breaks TLS certificate verification for **Go programs** inside the sandbox when using `httpProxyPort`.

## Root Cause

Go on macOS uses the Security framework directly (`SecTrustSettingsCopyTrustSettings`) to verify certificate trust, ignoring `SSL_CERT_FILE`. Without access to `trustd.agent`, Go programs cannot verify that custom CA certificates are trusted.

## Reproduction

1. Configure a MITM proxy with a custom CA
2. Add the CA to macOS Keychain with SSL trust
3. Set `httpProxyPort` to route traffic through the proxy
4. Run `gh api user` inside the sandbox

**Result:** TLS handshake fails - Go cannot verify the proxy's certificate is trusted.

**Verification:** Inside sandbox, `security dump-trust-settings` returns no results, while outside sandbox it correctly shows the trusted CA.

## Why PR #108's testing missed this

The tested tools (curl, wget, git, npm, pip, cargo, python, ruby, node) all use OpenSSL or respect `SSL_CERT_FILE`. None are Go programs.

Go programs on macOS are a special case:
- `crypto/tls` → `x509.SystemCertPool()` → Security framework → requires `trustd.agent`

## Affected tools

Any Go CLI that makes HTTPS requests:
- `gh` (GitHub CLI)
- `gcloud`
- `terraform`
- `kubectl`
- `hugo`
- etc.

## Suggested fix

Re-add `com.apple.trustd.agent` to the mach-lookup allowlist, at least when `httpProxyPort` is configured. This service only provides read access to certificate trust settings - it doesn't expose secrets or keychain items.

## Environment

- macOS 15.7.3 (Sequoia)
- sandbox-runtime 0.0.34+
- Go 1.24.8
- gh 2.82.0

ddworken:
Thank you for flagging! https://github.com/anthropic-experimental/sandbox-runtime/pull/120 will fix this with a new option to enable the older behavior.

connorshea:
To be clear with this specific option and it's current description, I don't believe this issue is exclusive to users with a proxy?

[...]

PR #120, opened 2026-02-05
Add enableWeakerNetworkIsolation config option for Go TLS support

ddworken:
Adds a new `enableWeakerNetworkIsolation` config option (macOS only, default: false) that conditionally re-enables access to `com.apple.trustd.agent` in the sandbox profile.

This is needed for Go programs (`gh`, `gcloud`, `terraform`, `kubectl`, etc.) that use the macOS Security framework for TLS certificate verification. Without `trustd.agent` access, Go programs cannot verify custom CA certificates (e.g. when using `httpProxyPort` with a MITM proxy).

This option is off by default. When enabled, it relaxes network isolation by allowing IPC to the trustd service which could be used as a data exfiltration vector.

Fixes #118


ollie-anthropic (approved):
thank you!!

PR #126, opened 2026-02-09
Re-introduce non-existent deny path protection with mount point cleanup

ddworken:
## Summary

Re-introduces the non-existent deny path protection from PR #80 (reverted in PR #91) with proper cleanup of bwrap mount point artifacts.

## Problem

When a deny path doesn't exist yet on the filesystem, a sandboxed process can `mkdir -p` + write to it, bypassing the deny rule. PR #80 fixed this by mounting `/dev/null` at the first non-existent component. However, bwrap creates empty files on the host as mount points for these binds, and they persisted after bwrap exited — causing ghost dotfiles (`.bashrc`, `.gitconfig`, etc.) to appear in the working directory (#85).

## Solution

- Re-introduce `findFirstNonExistentComponent()` and the `/dev/null` mounting logic
- Track all mount points created for non-existent deny paths in a module-level `Set<string>`
- Add `cleanupBwrapMountPoints()` which removes tracked mount points that are still empty files
- Add `SandboxManager.cleanupAfterCommand()` — a lightweight per-command cleanup API for callers to invoke after each sandboxed command exits
- Call `cleanupAfterCommand()` from `reset()` and on process exit as safety nets
- Call `cleanupAfterCommand()` from the `srt` CLI's child exit handler

## Safety

`cleanupBwrapMountPoints()` only removes files that are:
1. Tracked (added during `generateFilesystemArgs`)
2. Still exist
3. Are regular files with size 0

If something else wrote real content to the path, it is preserved.

## Test plan

- [x] Build passes (`tsc --noEmit`)
- [x] Existing tests pass (37/37 in mandatory-deny-paths)
- [x] New integration tests for:
  - Blocking non-existent file creation (parent exists)
  - Blocking non-existent file creation (parent also doesn't exist)
  - Blocking deeply nested non-existent paths
  - Cleanup removes mount point artifacts
  - Cleanup removes multiple mount points from a single command
  - Cleanup preserves files with real content
  - Cleanup is safe to call with no mount points (no-op)
  - End-to-end: no ghost dotfiles after command + cleanup cycle

Fixes #85


src/sandbox/linux-sandbox-utils.ts
```diff
@@ -230,11 +307,57 @@ function registerSeccompCleanupHandler(): void {
...
+ * Safe to call at any time — it only removes files that were tracked during
+ * generateFilesystemArgs() and skips any that no longer exist.
+ */
+export function cleanupBwrapMountPoints(): void {
+  for (const mountPoint of bwrapMountPoints) {
+    try {
+      // Only remove if it's still the empty file/directory bwrap created.
+      // If something else has written real content, leave it alone.
+      if (fs.existsSync(mountPoint)) {
+        const stat = fs.statSync(mountPoint)
```
dylan-conway:
```suggestion
      const stat = fs.statSync(mountPoint)
```
we can rely on try/catch to catch the exception from `fs.statSync`

ddworken: Done.

dylan-conway (approved):
LGTM! Only concern is `/tmp/claude-empty-` paths are not cleaned up, but this might be fine because they're already in the temp folder

PR #127, opened 2026-02-09
fix: use wildcard in allowLocalBinding seatbelt rules for IPv6 dual-stack compatibility

dylan-conway:
## Problem

Gradle builds fail in sandbox mode with `allowLocalBinding: true` because the Gradle daemon cannot bind to TCP localhost:

```
java.net.SocketException: Operation not permitted
    at sun.nio.ch.Net.bind0(Native Method)
    at TcpIncomingConnector.accept()
```

## Root Cause

Modern Java (and other runtimes) create **IPv6 dual-stack sockets** by default via `ServerSocketChannel.open()`. When binding such a socket to `127.0.0.1`, the kernel internally represents the address as `::ffff:127.0.0.1` (IPv4-mapped IPv6 address).

macOS Seatbelt's `(local ip "localhost:*")` filter resolves `localhost` to `127.0.0.1` and `::1`, but does **not** match `::ffff:127.0.0.1`. Seatbelt only supports two host values in IP filters: `localhost` and `*` — there is no way to specify `::ffff:127.0.0.1` explicitly.

| Socket type | Bind address | Seatbelt sees | Matches `localhost`? |
|---|---|---|---|
| AF_INET (Python, explicit Java) | `127.0.0.1` | `127.0.0.1` | Yes |
| AF_INET6 (Java default) | `127.0.0.1` | `::ffff:127.0.0.1` | **No** |
| AF_INET6 | `::1` | `::1` | Yes |

## Fix

Change the seatbelt rules from `(local ip "localhost:*")` to `(local ip "*:*")`.

## Safety Analysis

Tested via `sandbox-exec` — `(local ip "localhost:*")` is already more permissive than its name suggests. It allows binding, listening, inbound, and outbound on **all** local addresses, not just loopback:

| Operation | `localhost:*` | `*:*` |
|---|---|---|
| Bind/listen `127.0.0.1` | Allowed | Allowed |
| Bind/listen `0.0.0.0` | Allowed | Allowed |
| Bind/listen LAN IP | Allowed | Allowed |
| Bind/listen `::1` / `::` | Allowed | Allowed |
| Outbound TCP to internet | Allowed | Allowed |
| Accept inbound from LAN | Allowed | Allowed |
| Bind `::ffff:127.0.0.1` | **Blocked** | **Allowed** |
| Bind `::ffff:<LAN_IP>` | **Blocked** | **Allowed** |

The only new addresses `*:*` permits are the `::ffff:x.x.x.x` (IPv4-mapped IPv6) family. These are equivalent to their unmapped counterparts — every operation achievable via `::ffff:<addr>` can already be done via `<addr>` directly under `localhost:*`. The change grants no new network capability.

Verified via `sandbox-exec` testing:
- Java dual-stack IPC (bind + connect + accept): **works**
- Internet access: **blocked**
- Python IPv4 IPC: **works**
- Without `allowLocalBinding`: **blocked**

Fixes: https://github.com/anthropics/claude-code/issues/18545


ddworken (approved):
If it is practical, it would be cool to add some tests for this!

PR #135, opened 2026-02-15
fix: version reported by `srt --version`

MarlzRana:
The version reported by `srt --version` is incorrect. 

This makes the brew formula look peculiar (see the test): 

```rb
class SandboxRuntime < Formula
  desc "Lightweight sandboxing tool for filesystem and network restrictions"
  homepage "https://github.com/anthropic-experimental/sandbox-runtime"
  url "https://registry.npmjs.org/@anthropic-ai/sandbox-runtime/-/sandbox-runtime-0.0.37.tgz"
  sha256 "0ee00dbedeeb05c46e6c320d3d33de444cbafa2dd3345cafc15b06f549f65fa0"
  license "Apache-2.0"

  depends_on "node"

  on_linux do
    depends_on "bubblewrap"
  end

  def install
    system "npm", "install", *std_npm_args
    bin.install_symlink libexec.glob("bin/*")
  end

  test do
    assert_match "1.0.0", shell_output("#{bin}/srt --version")
    assert_match "Run commands in a sandbox", shell_output("#{bin}/srt --help")
  end
end
```

Let's fix that! 😄 

Fixes issue: https://github.com/anthropic-experimental/sandbox-runtime/issues/119


src/cli.ts
```diff
@@ -42,7 +46,7 @@ async function main(): Promise<void> {
     .description(
       'Run commands in a sandbox with network and filesystem restrictions',
     )
-    .version(process.env.npm_package_version || '1.0.0')
```
MarlzRana, line 45: `process.env.npm_package_version` is null when you do not invoke it via `npm` (e.g `srt echo "Hello world!"`).

PR #138, opened 2026-02-18
security: warn and skip symlink write paths pointing outside boundaries

ddworken:
## Problem

bwrap follows symlinks when doing bind mounts. If a user configures an `allowWrite` path that is a symlink pointing to an unexpected location, that target location becomes writable.

**Example attack:**
1. Attacker creates symlink: `./src -> /etc`
2. User configures `allowWrite: ['./src']`
3. bwrap does `--bind ./src ./src` and follows the symlink
4. `/etc` becomes writable through the symlink

## Fix

This change:
- Detects when a write path is a symlink pointing outside expected boundaries (using existing `isSymlinkOutsideBoundary` check)
- Prints a warning to inform the user
- Skips the path instead of making the unexpected target writable

## Testing

Unit tests (5 tests):
- Verifies normal non-symlink write paths are included in bwrap args
- Verifies symlink write paths pointing outside boundaries are skipped
- Verifies legitimate paths are kept while symlink paths are skipped
- Verifies broken symlinks are handled gracefully
- Verifies symlinks resolving within the same directory are allowed

Integration tests (4 tests):
- Verifies writes through symlink to protected directory are blocked
- Verifies normal writes to non-symlink paths still work
- Verifies writes through symlink pointing to `/etc` are blocked
- Verifies writes through symlink pointing to parent directory are blocked

All tests pass: `bun test ./test/sandbox/symlink-write-path.test.ts` (9 pass)


src/sandbox/linux-sandbox-utils.ts
```diff
@@ -672,6 +673,28 @@ async function generateFilesystemArgs(
         continue
       }
 
+      // Check if path is a symlink pointing outside expected boundaries
+      // bwrap follows symlinks, so --bind on a symlink makes the target writable
+      // This could unexpectedly expose paths the user didn't intend to allow
+      try {
+        const resolvedPath = fs.realpathSync(normalizedPath)
+        if (
+          resolvedPath !== normalizedPath &&
```
dylan-conway: during testing I ran into an issue where `normalizedPath` has a trailing slash and realpathSync returns it without one so the comparison would fail. I think we need to trim trailing slashes here

dylan-conway (changes requested):
One comment

src/sandbox/linux-sandbox-utils.ts
ddworken: Good catch! Fixed.

dylan-conway (approved):

PR #140, opened 2026-02-18
fix: allow Unix domain socket creation in network-restricted sandbox

thomasballinger:
## Problem

When `allowedDomains` is set, the sandbox enters restricted network mode. The previous implementation used `(allow network* (subpath "/"))` to allow Unix sockets, but `socket(AF_UNIX, SOCK_STREAM, 0)` is a `system-socket` operation that doesn't reference a filesystem path, so `(subpath ...)` can't match it.

This caused Gradle (`FileLockContentionHandler`), Docker, and other tools that create Unix domain sockets to fail with:
```
java.net.SocketException: Operation not permitted
```

Both `allowAllUnixSockets: true` and `allowUnixSockets: ["/path"]` were broken when `allowedDomains` was configured.

## Fix

Replace the single catch-all `(allow network* (subpath ...))` rule with three explicit Seatbelt operations:

1. `(allow system-socket (socket-domain AF_UNIX))` — for `socket()` creation (no path context)
2. `(allow network-bind (local unix-socket ...))` — for `bind()` operations
3. `(allow network-outbound (remote unix-socket ...))` — for `connect()` operations

This properly separates the socket creation syscall (which has no path context) from the bind/connect operations (which reference paths).

## Testing

Added integration tests that verify:
- `allowAllUnixSockets: true` — creates, binds, and connects a Unix socket
- `allowUnixSockets: ["/specific/path"]` — same with path-scoped permissions
- Neither setting — socket bind is correctly blocked


dylan-conway (changes requested):
Looks good, but mind deleting the `bun.lock`? We can do a full migration in a separate PR

dylan-conway (approved):

PR #187, opened 2026-03-27
Add upstream/parent HTTP proxy support to sandbox

MarshallOfSound:
## Summary

Allows the sandbox to run behind a corporate HTTP proxy. When configured (via `network.parentProxy` or the standard `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env vars), the sandbox's HTTP and SOCKS proxies tunnel outbound traffic through the upstream proxy instead of connecting directly.

- **`parent-proxy.ts`** — new module handling config resolution, `NO_PROXY` matching (hostname suffix + IPv4/IPv6 CIDR, golang `httpproxy` semantics), and a `CONNECT`-tunnel helper that returns a raw socket piped through the parent
- **`http-proxy.ts`** — direct-path `CONNECT` and plain HTTP requests now chain through the parent; loopback and `NO_PROXY` matches still connect directly
- **`socks-proxy.ts`** — custom connection handler routes SOCKS requests through the parent's HTTP `CONNECT`
- **`sandbox-config.ts`** / **`sandbox-manager.ts`** — new `network.parentProxy` schema field; resolved once at init before child env vars are shadowed

MITM-proxied domains are unaffected — only the direct-connect path is tunnelled.

## Test plan

- [x] Unit tests: config resolution (explicit > env, lowercase vars, HTTPS→HTTP fallback)
- [x] Unit tests: `NO_PROXY` matching — exact, subdomain, leading-dot, `*.`, `*`, CIDR v4, loopback-always-bypass, case-insensitive, port stripping
- [x] E2E: `CONNECT` through a recording parent proxy → verify origin reached and parent saw the `CONNECT`
- [x] E2E: loopback destination bypasses parent even with empty `NO_PROXY`
- [ ] Manual: run behind a real corporate proxy


src/sandbox/parent-proxy.ts
```diff
@@ -0,0 +1,519 @@
...
+        : netConnect(proxyPort, proxyHost),
+  })
+}
+
+// ---------------------------------------------------------------------------
+// Utilities
+// ---------------------------------------------------------------------------
+
+export function proxyAuthHeader(proxyUrl: URL): string | undefined {
+  if (!proxyUrl.username) return undefined
```
dylan-conway:
```suggestion
  if (!proxyUrl.username && !proxyUrl.password) return undefined
```
very unlikely but I believe it's possible for only the password to be set

src/sandbox/parent-proxy.ts
```diff
@@ -0,0 +1,519 @@
...
+import { BlockList, connect as netConnect, isIP } from 'node:net'
+import { connect as tlsConnect } from 'node:tls'
+import { URL } from 'node:url'
+import { logForDebugging } from '../utils/debug.js'
+
+export interface ParentProxyConfig {
+  http?: string
+  https?: string
+  noProxy?: string
+}
```
dylan-conway:
```suggestion
```
should be replaced with `ParentProxyConfig` import from sandbox-config.ts

dylan-conway (approved):
LGTM

Issue #10, opened 2025-10-24
Cannot express “workspace-only filesystem access” with current sandbox rules


smolcompute:
## **Expected Sandbox Possible Setups**

1. Allow all Read|Edit|Write inside the workspace (CWD). 
2. Hard-block all filesystem access outside the workspace (CWD). 

AND 

1. Set global (user) deny in ~/.claude/settings.json
2. Set project specific ask in project/.claude/settings.local.json

## **Actual**

Because deny trumps everything else and the rules don't allow for exceptions, any deny hierarchically higher than the CWD will also block the CWD ...

Likewise, the project's settings file is supposed to have precedence on the global user settings file, but because deny trumps asks or allow, if you set deny globally, you can't set project specific ask or allow ...  global deny will trump project specific rules ...

With precedence Deny > Ask > Allow and no exception/negation syntax, any deny that matches a parent scope (e.g., ~/**) also blocks the workspace subtree when the workspace lives under it. A parent-escape rule like ../** does not catch absolute paths (/Users/...), ~, or symlink targets. “Ask” is not a boundary because user consent allows the read.

## Why this cannot be modeled today

1. No “except” rules; deny always wins across User + Workspace scopes.
2. No predicate like “inside workspace realpath,” so absolute or ~ paths bypass ../**.
3. Symlinks can point outside the workspace; string patterns don’t resolve targets.
4. A blanket deny(//**), which would seem to help, also kills absolute paths to files inside the workspace and cannot be overridden.

## **Workarounds (insufficient)**
- **No workaround possible for global deny rules cancelling project specific allow or ask rules.** 

- Keep the workspace under ~ and continually maintain a growing deny list for home subtrees + system roots. Fragile and high-maintenance.

- Move the workspace outside ~ (e.g., /Users/Shared/CWD) so you can deny ~/** globally and explicitly allow the workspace path. Closer to a dev container, but requires relocation, manually denying all other paths but  /Users/Shared, and breaks running claude code on other projects located in ~.

## **Request**

Add a first-class way to confine file access to the workspace, for example:
- A boolean workspaceOnlyFs: true that restricts all file ops to the workspace realpath.
- Or a path predicate insideWorkspace() evaluated on the resolved realpath.
- Or scoped precedence that lets a Workspace allow override a broader User deny only for the workspace subtree.
- Optionally resolve symlinks before policy evaluation to prevent path tunneling.

richardscarrott:
[...]

PR #190, opened 2026-03-30
fix: allow filesystem root traversal when denyRead includes '/'

poteat:
## Summary

Two fixes for Linux/macOS sandbox filesystem rules.

**Root-level denyRead carve-outs** (follow-up to #166, fixes #10). `denyRead: ['/']` with `allowRead: [<project>]` denied everything on both platforms. On macOS, `(subpath "/")` denies the root inode itself and no re-allow subpath covers it, so dyld SIGABRTs before exec; we emit a literal-root allow so path traversal works. On Linux, `--tmpfs /` wiped every prior mount and the carve-out prefix check compared against `'//'`; we expand a root deny into its children instead. While here we also restore write binds that a denyRead tmpfs over an ancestor wipes (a pre-existing issue affecting `denyRead: ['~']` too), and reorder denyWrite binds last so e.g. `.git/hooks` protection survives.

**Duplicate denyWrite bwrap failure** (re anthropics/claude-code#38375). Two config entries that converge to the same path only after normalization produce a duplicate `--ro-bind /dev/null` mount; the second one hits a char device and bwrap's `ensure_file()` falls through to `creat()` on a read-only mount. We dedup post-normalization.

## Links

- Fixes #10
- Follow-up to #166
- re anthropics/claude-code#38375

## Test Plan

- [ ] macOS: `sandbox-exec -p '(version 1)(allow default)(deny file-read* (subpath "/"))(allow file-read* (literal "/"))(allow file-read* (subpath "/bin"))(allow file-read* (subpath "/usr"))(allow file-read* (subpath "/System"))' /bin/echo hi` exits 0 (was SIGABRT without the literal rule)
- [ ] Linux CI passes the new `allow-read.test.ts` root-deny cases (unverified locally, no bwrap here)
- [ ] Linux CI passes the new `wrap-with-sandbox.test.ts` dedup case


src/sandbox/linux-sandbox-utils.ts
```diff
@@ -845,24 +873,40 @@ async function generateFilesystemArgs(
...
           if (!fs.existsSync(allowPath)) {
             logForDebugging(
               `[Sandbox Linux] Skipping non-existent read allow path: ${allowPath}`,
             )
             continue
           }
+          // Skip if already re-bound as writable above
+          if (
+            allowedWritePaths.some(
+              w => allowPath === w || allowPath.startsWith(w + '/'),
```
dylan-conway: **Skip-check too broad — `allowRead` dropped when `allowWrite` is an *ancestor* of `denyRead` (regression)**

This checks against **all** `allowedWritePaths`, but the re-bind loop at 882-889 only restores writes that are **under** the current `denyRead` (`writePath.startsWith(denySep)`). When a write path is *above* the `denyRead`, it is correctly not re-bound (the tmpfs is below it, doesn't wipe it) — yet this skip still fires because `allowPath` is transitively under that ancestor write.

**Trigger** — matches the CCR production topology (`allowWrite:~`, `denyRead:~/.ssh`):
```
allowWrite: ['/home/user']
denyRead:   ['/home/user/.ssh']
allowRead:  ['/home/user/.ssh/known_hosts']
```
→ `--bind /home/user` → `--tmpfs /home/user/.ssh` → re-bind loop: `/home/user` not under `.ssh/`, nothing re-bound → skip check: `known_hosts` IS under `/home/user/` → **skipped**. `known_hosts` sits in empty tmpfs, unreadable.

Pre-PR (fd74a3f) had no skip — it ro-bound unconditionally, so this is a regression. The comment at 902 ("Skip if already re-bound as writable above") confirms the intent was narrow.

**Fix:** narrow the predicate to writes actually re-bound in this iteration:
```ts
allowedWritePaths.some(w =>
  (w.startsWith(denySep) || w === normalizedPath) &&
  (allowPath === w || allowPath.startsWith(w + '/'))
)
```

poteat: Good catch, thanks — fixed in f16dbed. Narrowed the predicate to only skip when `w` itself is under `denySep` (i.e. was actually re-bound just above). Added a regression test in `allow-read.test.ts` with `allowWrite: [TEST_BASE_DIR]` as ancestor of the deny dir.

dylan-conway (approved):

Issue #171, opened 2026-03-13
Cannot make directory using allowRead writeable (Linux,git HEAD)


JanWielemaker:
Using GIT tag v0.0.42 version on Fedora Linux 43, I'm trying to limit access to just a project dir using

```
  "filesystem": {
    "denyRead": ["/home", "/etc"],
    "allowRead": ["~/src/swipl-devel", "~/3rdparty"],
    "allowWrite": ["~/src/swipl-devel"],
    "denyWrite": []
  },
```

While this gets the expected read permissions, I cannot write in `~/src/swipl-devel`.   Going there:

```
> srt -d touch x
touch: cannot touch 'x': Read-only file system
```

As you see, `-d` emits nothing as well. 

allowWrite works fine if I do not use allowRead, but only deny read access to some directories.

kulla:
I have the same issue and investigated it. At https://github.com/kulla/2026-03-18-testing-srt/tree/main/03-bug-order-write-and-read one can find a minimal example.

It seems that in calling `bwrap` by `srt`, first all write bindings are passed and only afterwards the arguments for denying / allowing read operations. In my example I found a `bwrap` call like:

```bash
bwrap  ... \\
  --bind /home/kulla/workspace/kulla/2026-03-18-testing-srt/03-bug-order-write-and-read/file.txt /home/kulla/workspace/kulla/2026-03-18-testing-srt/03-bug-order-write-and-read/file.txt \\
  --tmpfs /home
  ..
```

Thus the `--tmpfs /home` seems to overwrite the write access to `file.txt` since it is located in a subdirectory of `/home`.

Benjamin-Annunciation-Interactive:
Seeing the same fundamental issue on macOS (Seatbelt).
                                                                                                                                                                                                         
  Setup:          
  // ~/.claude/settings.json                                                                                                                                                                             
  "sandbox": {              
    "enabled": true,                                                                                                                                                                                     
    "filesystem": { 
      "denyRead": ["~/"]                                                                                                                                                                                 
    }                   
  }  
   
  // .claude/settings.local.json (project level)
  "sandbox": {                                                                                                                                                                                           
    "filesystem": {
      "allowRead": ["."],                                                                                                                                                                                
      "allowWrite": ["."]
    }                    
  }  
   
Expected: Full read/write/delete access within the project directory (which is under ~/), with reads denied everywhere else.                                                                           
                                                                                                                                                                                                         
[...]

PR #198, opened 2026-03-31
test: verify rm in allowWrite under denyRead ancestor (follow-up to #170)

poteat:
Adds integration tests for the fix in #170. Covers the three properties from #171 (Benjamin's config shape): rm works in allowWrite under a denyRead ancestor, stays blocked outside allowWrite, stays blocked in denyWithinAllow. Plus the Linux analogue from #190.


bogini (approved):
stamped 🔨

dylan-conway (approved):

PR #199, opened 2026-04-02
Bake BPF filter into apply-seccomp, build in CI

dylan-conway:
The unix-block BPF filter bytes are now generated as a C header at build
time and compiled into apply-seccomp. The separate .bpf file and all the
TS path-resolution, caching, and cleanup machinery around it are gone.

## What's new

- `vendor/seccomp/build.ts` — compiles the BPF generator, runs it for both
  x64 and arm64, writes the bytes into `unix-block-bpf.h`, then compiles
  apply-seccomp with that header `#include`d. Arch selection happens via
  `#if defined(__x86_64__) / #elif defined(__aarch64__)` in the header.
- `apply-seccomp.c` takes one fewer argv — just the command now, no filter
  path. The filter bytes are a `static const unsigned char[]` in the binary.

## Binaries no longer committed

`release.yml` builds on both an x64 and an arm64 runner, uploads each as
an artifact, and the publish job downloads both into
`vendor/seccomp/{x64,arm64}/` before `npm publish`. Tarball layout is
unchanged.

`integration-tests.yml` runs `npm run build:seccomp` before tests on
Linux jobs.

## Deleted

TS: `generateSeccompFilter`, `cleanupSeccompFilter`, `getPreGeneratedBpfPath`,
`findBpfPath`, `bpfPathCache`, `generatedSeccompFilters`, `seccompFilterPath`
threading, `bpfPath` from `SeccompConfigSchema` and `checkLinuxDependencies`,
`hasSeccompBpf` from `LinuxDependencyStatus`.

Scripts: the Docker-based `build-seccomp-binaries.sh`.


blois (approved):

szwang (commented):
Reviewed for #claude-code-stamps. Clean simplification — −864/+215, removes a whole layer of TS path-resolution machinery by baking the BPF bytes into the loader binary. A few notes:

**Correctness**
- `seccomp-unix-block.c` was correctly extended to accept an optional target arch arg (`seccomp_arch_remove(NATIVE)` + `seccomp_arch_add(target)`), so the single-host build can emit both filters. Verified the call site in `build.ts` matches: `run([gen, tmp, target])`.
- `_Static_assert(sizeof(unix_block_bpf) % sizeof(struct sock_filter) == 0)` is a nice compile-time replacement for the old runtime `filter_size % 8 != 0` check.
- The `(struct sock_filter *)unix_block_bpf` cast drops `const`, but `prctl(PR_SET_SECCOMP)` doesn't write to the buffer (kernel copies it), so the `.rodata` placement is safe.
- `registerExitCleanupHandler` still has live callers for the bwrap mount-point cleanup path — not orphaned.

**Release pipeline**
- `download-artifact` doesn't preserve the executable bit; the `chmod +x vendor/seccomp/*/apply-seccomp` step handles it. ✓
- `if-no-files-found: error` on the upload prevents a silent empty-artifact publish. ✓
- Nit: `release.yml` still has an explicit `npm run clean && npm run build` step before `npm publish`, but `prepublishOnly` now does the same thing again (plus the vendor copy). The explicit step is wasted work — harmless, but could be dropped.

**API surface**
- Removed exports (`generateSeccompFilter`, `cleanupSeccompFilter`, `getPreGeneratedBpfPath`, `hasSeccompBpf`, `bpfPath`) are not re-exported from `src/index.ts` and have no consumers in claude-cli-internal. `SeccompConfigSchema` is non-strict so any stale `bpfPath` in a downstream config gets silently stripped rather than erroring. Safe break.

**CI**
- The `linux/arm64` failure is `should enforce network restrictions across protocols and ports` timing out at 5001ms — that's a network-reachability test at L700 of `integration.test.ts`, untouched by this PR. `linux/x86-64` passed the same suite with the freshly-built binary, so the seccomp changes themselves are fine. Looks like an arm64 runner network flake; worth a re-run.

LGTM.

ddworken (approved):

Issue #151, opened 2026-02-25
Cannot use `gcloud` CLI tool due to unsupported `CLOUDSDK_PROXY_TYPE=https`


kwohlfahrt:
I am running a sandbox with the following settings:

```settings
{
  "filesystem": {
    "allowWrite": ["~/.config/gcloud/logs"],
    "denyWrite": []
  },
  "network": {
    "allowedDomains": [],
    "deniedDomains": []
  }
}
```

When I try to run the `gcloud storage ls` command, I get the error below:

```
> npm run-script -- srt -s ./srt-settings.json gcloud storage ls gs://vantq-argo/moleculearn/runs/kai-debug/profile-multinode-2025-10-31-11-41-59/profiling/

> srt
> srt -s ./srt-settings.json gcloud storage ls gs://vantq-argo/moleculearn/runs/kai-debug/profile-multinode-2025-10-31-11-41-59/profiling/

ERROR: (gcloud.storage.ls) Invalid value for property [proxy/type]: The proxy type property value [https] is not valid. Possible values: [http, http_no_tunnel, socks4, socks5].
```

This seems to be due to the `CLOUDSDK_PROXY_TYPE=https` env var set [here](https://github.com/anthropic-experimental/sandbox-runtime/blob/ef4afdef4d711ba21a507d7f7369e305f7d3dbfa/src/sandbox/sandbox-utils.ts#L376). The supported values are `http`, `http_no_tunnel`, `socks4`, and `socks5` ([docs](https://docs.cloud.google.com/sdk/docs/proxy-settings#proxy_configuration)).

I initially filed anthropics/claude-code#28579, but have since confirmed it occurs directly with the sandbox.

danielorbach:
Some senseless LLM probably set this to `https` (because `gcloud` clearly never supported it) yet nobody on the official team of maintainers got to change this 1 byte 🐡 

@dylan-conway @cirospaciari @poteat please fix this

danielorbach:
Thank you for the fix @dylan-conway 🎉 

Can you explain how changes in this repo land in Claude Code? So we know when to expect this fix in users' workstations 🙏

PR #237, opened 2026-05-05
fix(sandbox): set CLOUDSDK_PROXY_TYPE=http (was invalid "https")

dylan-conway:
Fixes #151.

`gcloud` rejects `CLOUDSDK_PROXY_TYPE=https`:

```
ERROR: (gcloud.storage.ls) Invalid value for property [proxy/type]: The proxy type property value [https] is not valid. Possible values: [http, http_no_tunnel, socks4, socks5].
```

Per [Google's proxy docs](https://docs.cloud.google.com/sdk/docs/proxy-settings), `proxy/type` only accepts `http`, `http_no_tunnel`, `socks4`, or `socks5`. The setting names the protocol the *proxy* speaks, not the traffic it tunnels — the sandbox's HTTP CONNECT proxy carries TLS traffic just fine, so `http` is correct here (same reason `HTTPS_PROXY` is already set to `http://localhost:<port>` a few lines up).

Also adds a unit test for `generateProxyEnvVars` covering this.


ig-ant (approved):

chrislloyd (approved):
stamped 🔨

Issue #157, opened 2026-02-27
Positional CLI arguments containing spaces are broken by commandArgs.join(' ')


tesdal:
## Description

When using the CLI in positional argument mode (not `-c`), arguments containing spaces are broken because `commandArgs.join(' ')` on line 150 of `src/cli.ts` destroys argument boundaries.

The joined string is then passed through `bash -c <command>` where the shell re-parses it, splitting on whitespace — so `"hello world"` becomes two separate tokens `hello` and `world`.

## Reproduction

```bash
# Expected: prints "hello world\n"
# Actual: prints "hello\nworld\n" (two separate arguments)
srt /usr/bin/printf '%s\n' "hello world"

# Real-world example:
# Expected: creates PR with title "docs: add architecture guide"
# Actual: gh errors with unknown arguments ["add" "architecture" "guide"]
srt gh pr create --title "docs: add architecture guide"
```

## Root Cause

`src/cli.ts` line 150:
```typescript
command = commandArgs.join(' ')
```

The `-c` flag path is **not affected** — it receives a single string and passes it through verbatim.

## Impact

Any CLI tool invoked through `srt` with positional arguments that contain spaces will malfunction. This commonly affects:
- `gh pr create --title "..."` / `--body "..."`
- `git commit -m "message with spaces"`
- Any command with file paths containing spaces

## Suggested Fix

Replace `commandArgs.join(' ')` with `shellquote.quote(commandArgs)`. The `shell-quote` package is already a production dependency, used in `macos-sandbox-utils.ts` and `linux-sandbox-utils.ts`.

PR #239, opened 2026-05-05
fix(cli): shell-quote positional args instead of join(" ")

dylan-conway:
Fixes #157.

In positional mode (`srt cmd arg1 arg2`), argv is reassembled into a string and executed via `bash -c`. Reassembling with `commandArgs.join(" ")` discards argument boundaries, so any arg containing whitespace or shell metacharacters is re-tokenised by bash:

```
srt printf "%s\n" "hello world"
# before: hello\nworld\n  (two args)
# after:  hello world\n
```

```
srt gh pr create --title "docs: add architecture guide"
# before: gh errors on stray args "add" "architecture" "guide"
```

`shell-quote` is already a production dependency (used by the macOS/Linux wrappers). The `-c` path is unchanged — it takes a single string and is meant to be shell-parsed.

Adds tests using `printf "%s\n"` (one line per arg) to make the boundary loss observable, plus a metacharacter-preservation case.


ig-ant (approved):

chrislloyd (approved):
stamped 🔨

Issue #141, opened 2026-02-19
TMPDIR mismatch: sandbox runtime reads `CLAUDE_TMPDIR` but Claude Code sets `CLAUDE_CODE_TMPDIR`


arakashic:
## Summary

The sandbox runtime reads `CLAUDE_TMPDIR` to determine the temp directory path, but Claude Code now sets `CLAUDE_CODE_TMPDIR` instead. Since `CLAUDE_TMPDIR` is not set, the runtime falls back to `/tmp/claude`, which does not exist on disk. This causes child processes (e.g. cmake, nvcc) that respect `TMPDIR` to fail when writing temp files.

## Environment

Reproduced on both Linux and macOS:

**Linux:**
- Ubuntu 24.04 (kernel 6.8.0-90-generic)
- Claude Code CLI (claude-opus-4-6)
- CUDA 13.1.80

**macOS:**
- macOS (UID 501)
- Claude Code CLI

## Steps to Reproduce

1. Run any command via Claude Code's sandboxed Bash that invokes a tool writing to `TMPDIR` (e.g. `cmake --preset dev` on a CUDA project)
2. The command fails because `TMPDIR` points to a non-existent directory

## Observed Behavior

**Linux (Ubuntu 24.04, UID 1001):**

```
$ env | grep -i claude | grep -i tmp
TMPDIR=/tmp/claude
CLAUDE_CODE_TMPDIR=/tmp/claude-1001
TMPPREFIX=/tmp/claude-1001/zsh

$ ls /tmp/ | grep claude
claude-1001/
# Note: /tmp/claude does NOT exist

$ cmake --preset dev
nvcc fatal : Could not open output file '/tmp/claude/tmpxft_0000000c_0000000a'

# Workaround: override TMPDIR
$ TMPDIR=/tmp/claude-1001 cmake --preset dev
-- Configuring done (6.1s)
-- Generating done (1.8s)
```

**macOS (UID 501):**

```
TMPDIR=/tmp/claude
CLAUDE_CODE_TMPDIR=/tmp/claude-501
TMPPREFIX=/tmp/claude-501/zsh
```

The sandbox write allowlist references `/private/tmp/claude-501/` (the macOS canonical path), which matches `CLAUDE_CODE_TMPDIR` but not `TMPDIR`.

## Root Cause

[...]

PR #240, opened 2026-05-05
fix(sandbox): read CLAUDE_CODE_TMPDIR for TMPDIR (in addition to CLAUDE_TMPDIR)

dylan-conway:
Fixes #141.

`generateProxyEnvVars()` derives the sandboxed `TMPDIR` from `CLAUDE_TMPDIR`, but callers now set `CLAUDE_CODE_TMPDIR` instead. With only the old name checked, `TMPDIR` falls back to the hardcoded `/tmp/claude` — which is not created and is not in the write allowlist — so tools that honour `TMPDIR` fail:

```
$ env | grep -i tmp
TMPDIR=/tmp/claude
CLAUDE_CODE_TMPDIR=/tmp/claude-1001

$ cmake --preset dev
nvcc fatal : Could not open output file '/tmp/claude/tmpxft_0000000c_0000000a'
```

Read `CLAUDE_CODE_TMPDIR` first, keep `CLAUDE_TMPDIR` as a fallback for backwards compatibility, then default to `/tmp/claude` as before.


ig-ant (approved):

chrislloyd (approved):
stamped 🔨

Issue #156, opened 2026-02-26
macOS: checkDependencies() requires ripgrep (rg) but macOS sandbox doesn't use it


rock3r:
## Summary

`checkDependencies()` in `sandbox-manager.js` requires `rg` (ripgrep) on **all** platforms, including macOS. However, the macOS sandbox (`wrapCommandWithSandboxMacOS`) does **not** use ripgrep — it uses `sandbox-exec` with regex patterns generated by `globToRegex()`. The macOS code even has the comment:

```js
// Combine user-specified and mandatory deny patterns (no ripgrep needed on macOS)
```

This means `SandboxManager.initialize()` throws on macOS if `rg` is not installed, even though `rg` is never actually used there.

## Impact

If a consumer calls `initialize()` on a macOS system without `rg` on the PATH, the sandbox fails to initialize:

```
Error: Sandbox dependencies not available: ripgrep (rg) not found
```

This silently degrades to no OS-level sandboxing if the error is caught (as it typically would be), leaving only the consumer's own permission layer as protection.

## Where the bug is

`sandbox-manager.js`, `checkDependencies()`:

```js
function checkDependencies(ripgrepConfig) {
    // ...
    const rgToCheck = ripgrepConfig ?? config?.ripgrep ?? { command: 'rg' };
    if (whichSync(rgToCheck.command) === null) {
        errors.push(`ripgrep (${rgToCheck.command}) not found`);  // ← fires on macOS too
    }
    if (platform === 'linux') {
        // Linux-specific checks (bwrap, socat)
    }
    return { errors, warnings };
}
```

The `rg` check runs unconditionally before the Linux-specific branch. It should be gated to Linux only, since:
- **Linux**: `rg` is used by `linuxGetMandatoryDenyPaths()` to scan for dangerous files (bubblewrap doesn't support globs)
- **macOS**: `sandbox-exec` profiles support regex/glob natively via `globToRegex()`, so no file scanning is needed

## Suggested fix

```js
if (platform === 'linux') {
    // Only check rg on Linux where it's actually used
    if (whichSync(rgToCheck.command) === null) {
        errors.push(`ripgrep (${rgToCheck.command}) not found`);
    }
    // ... other Linux checks
}
```

## Versions

Verified on both `0.0.26` and `0.0.39` — same behavior in both.

PR #241, opened 2026-05-05
fix(sandbox): only require ripgrep on Linux in checkDependencies()

dylan-conway:
Fixes #156.

`ripgrep` is used by `linuxGetMandatoryDenyPaths()` to expand glob deny patterns into concrete paths for bwrap. The macOS path never invokes it — seatbelt profiles take regex patterns directly (the existing `// no ripgrep needed on macOS` comment in `macos-sandbox-utils.ts` already says as much).

`checkDependencies()` was checking for `rg` unconditionally, so on a macOS host without `rg` on `PATH`, `SandboxManager.initialize()` throws

```
Sandbox dependencies not available: ripgrep (rg) not found
```

even though `rg` is never used there.

Moves the check inside the `platform === 'linux'` branch and adds tests covering both platforms.


ig-ant (approved):

chrislloyd (approved):
stamped 🔨

PR #246, opened 2026-05-05
fix: Allow network access while file system isolation is enabled.

dileepyavan:
fixes #253
## Summary

This PR adds support for running sandboxed commands with normal network access when a user explicitly opts in, while continuing to keep filesystem protections in place.

## Why

Some developer workflows need unrestricted network access, but still benefit from limiting what files a command can read or write. This change makes that use case easier to configure and clearer to document.

## What changed

- Adds a configuration option to allow normal network access when needed.
- Keeps filesystem sandboxing active even when network restrictions are turned off.
- Updates the README with guidance and examples for this mode.
- Adds test coverage to make sure the behavior stays consistent.


billziss-gh:
Please consider merging this. It is particularly useful in autoresearch style projects where one wants to control file system access but allow free network access for research.

PR #278, opened 2026-05-22
feat(windows): srt-win crate — group + wfp subcommands

ig-ant:
PR 1/8 of Windows sandbox support. Adds the standalone Rust helper binary that manages the discriminator group and the persistent WFP filter set; no TS integration yet.

## What this PR adds

**`vendor/srt-win/`** — Rust crate, binary `srt-win.exe`, with two noun subcommands:

- `srt-win group create|status|delete (--name <N> | --group-sid <S>)` — manage the local discriminator group (default `sandbox-runtime-net`). `status` emits `{"state":"absent|created-not-on-token|ready","sid":"S-1-…"}`.
- `srt-win wfp install|status|uninstall (--name <N> | --group-sid <S>) [--sublayer-guid <G>]` — manage the machine-wide persistent WFP filters. `status` emits `{"state":"absent|installed","filters":N}`.

`create`/`install`/`delete`/`uninstall` require elevation; `status` does not.

**`vendor/srt-win/ci/smoke.ps1`** + `cleanup.ps1` — full lifecycle smoke test, runnable locally on a Windows dev box.

**`.github/workflows/integration-tests.yml`** — `windows-latest` (x64) and `windows-11-arm` matrix legs: cargo build/test/clippy, then `smoke.ps1`.

## Filter design

One **machine-wide** filter set per sublayer — four filters at each of `FWPM_LAYER_ALE_AUTH_CONNECT_V4/V6` (8 total), keyed only on `<group_sid>`:

| # | Action | Weight | Condition |
|---|---|---|---|
| 0 | PERMIT | 0xF | `ALE_USER_ID` = `O:LSG:LSD:(D;;CC;;;<group_sid>)(A;;CC;;;WD)` |
| 1 | PERMIT | 0xE | `ALE_USER_ID` = `O:LSG:LSD:(A;;CC;;;<group_sid>)` |
| 2 | PERMIT | 0xD | `IP_REMOTE_ADDRESS` = `127.0.0.0/8` (v4) / `::1` (v6) |
| 3 | BLOCK  | 0x1 | `ALE_USER_ID` = `O:LSG:LSD:(A;;CC;;;WD)` |

WFP evaluates `ALE_USER_ID` `FWP_MATCH_EQUAL` by `AccessCheck`-ing the connecting token against the SD; the filter matches iff the check grants. Filter 0's SD has DENY-`<group>` before ALLOW-Everyone, so it grants only when the group is **absent** from the token — services, SYSTEM, and users not in the group all hit filter 0 → PERMIT, untouched. Filter 1 grants when the group is **enabled** (broker). A token with the group **deny-only** (sandboxed child) matches neither — deny-only SIDs match DENY ACEs (so filter 0 denies) but are ignored by ALLOW ACEs (so filter 1 doesn't grant) — and falls through to filter 2 (loopback, for the host proxy) or filter 3 (BLOCK).

`tests/sd_access_check_matrix.rs` proves that `AccessCheck` asymmetry directly — the 3×3 matrix of (enabled / deny-only / absent) tokens against the three SDs — without a live WFP engine.

## Why machine-wide

One filter set per machine instead of per-user. Enterprises add users to the group (locally or via domain GPO) and run `srt-win wfp install` once; no per-user WFP step, no need to know fenced users' SIDs at install time, no growth of the filter table as users are onboarded.

## Design notes

- Filters carry a `providerData` JSON tag (`{tool,kind}`); install/status/uninstall locate them by enumeration. **`wfp status` identifies filters by their `providerData` tag; filters installed by other tooling without the tag are not detected** — enterprise WFP rollouts must either use `srt-win.exe wfp install` or replicate the tag.
- v4 loopback permit is `FWP_V4_ADDR_AND_MASK 127.0.0.0/8`; v6 is exact `::1`.
- Sublayer GUID is a compile-time default, overridable via `--sublayer-guid`.
- No marker file — all status comes from live enumeration.

## Deliberately not in this PR

- `exec` subcommand (restricted token + job + spawn) — PR 2
- TS/`SandboxManager` integration — PR 3
- `acl` subcommand (filesystem deny) — PR 4–6
- Binary distribution via optional platform packages — PR 7
- README / docs — PR 8


vendor/srt-win/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "srt-win"
+version = "0.0.1"
+edition = "2021"
```
ig-ant: Why Rust edition 2021 and not edition 2024?

.github/workflows/integration-tests.yml
```diff
@@ -27,21 +27,156 @@ jobs:
...
+        shell: pwsh
+        run: |
+          cargo build  --release --manifest-path vendor/srt-win/Cargo.toml
+          cargo test   --release --manifest-path vendor/srt-win/Cargo.toml
+          cargo clippy --release --manifest-path vendor/srt-win/Cargo.toml --all-targets -- -D warnings
+
+      - name: Smoke-test srt-win group + wfp lifecycle (Windows)
+        if: matrix.os == 'windows'
+        shell: pwsh
+        run: |
```
ig-ant: For tests like this - I would prefer if they were pulled out into separate files which can be run locally, instead of being inline in the workflow yaml.

vendor/srt-win/Cargo.toml
ig-ant: Bumped to edition 2024 in 0c5f878. No fixups required — the crate has no `unsafe fn` (only `unsafe` blocks in safe fns, which 2024 leaves alone) and no `gen` identifiers. `cargo clippy --all-targets -- -D warnings` is clean on both the Windows target and the host.

.github/workflows/integration-tests.yml
ig-ant: Extracted to `vendor/srt-win/ci/smoke.ps1` (+ `cleanup.ps1`) in 0c5f878; runnable locally via `pwsh vendor/srt-win/ci/smoke.ps1 <path-to-srt-win.exe>`. The workflow steps now just call those scripts. smoke.ps1 carries the Run/J/MustFail helpers and writes $SRT_ALT_GUID to GITHUB_ENV so cleanup.ps1 can tear down the alternate sublayer on failure.

vendor/srt-win/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+path = "src/main.rs"
+
+[dependencies]
+anyhow = "1"
+clap = { version = "4", features = ["derive"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+
+[target.'cfg(windows)'.dependencies]
+windows = { version = "0.58", features = [
```
ig-ant: Can we use the newer version "0.62.2"?

ig-ant: Bumped to 0.62.2 in e832c93; 11 call-site fixups (nullable params -> `Option<T>`, `BOOL` moved to `windows::core`, `GUID::from(&str)` -> `try_from`).

dylan-conway (approved):

PR #285, opened 2026-05-28
feat(windows): single-source proxy env — TS owns generateProxyEnvVars, srt-win exec passthrough

ig-ant:
Follow-up to the network sandbox PR.

## What

Makes the TS layer the **single source** of the proxy environment on Windows, and turns `srt-win exec` into a verbatim environment passthrough.

- **TS (`windows-sandbox-utils.ts`)**: `wrapCommandWithSandboxWindows()` now returns `{ argv, env }`, where `env = { ...process.env, ...generateProxyEnvVars(httpPort, socksPort) }` — the same canonical builder macOS/Linux use (HTTP(S)_PROXY/ALL_PROXY both cases, NO_PROXY private-range list, plus the long tail: DOCKER_*, GRPC_PROXY, JAVA_TOOL_OPTIONS, …).
- **`SandboxManager.wrapWithSandboxArgv()`** returns `{ argv, env }` on every platform (on macOS/Linux `env` is the unchanged `process.env`, proxy is baked into the wrapped command as before) so callers spawn uniformly: `spawn(argv[0], argv.slice(1), { shell: false, env })`.
- **`srt-win exec` (launch.rs / main.rs)**: `--http-proxy` / `--socks-proxy` flags removed; `build_env_block()` forwards the broker's environment verbatim (sorted, unmodified values) plus `SANDBOX_RUNTIME_WIN_BROKER_PID`. No proxy synthesis, no stripping.
- `src/cli.ts` spawn site and `smoke-exec.ps1` updated for the new shape; `test/helpers/spawn.ts` gains an `env` passthrough option.

## Why

One place to maintain the proxy-env contract: every tool-specific variable added for macOS/Linux is automatically delivered to Windows children too, instead of a second hand-rolled list inside the Rust broker drifting out of sync. The WFP port-range fence is unchanged — env vars remain a compatibility courtesy, not the boundary.

## Test plan

- `test/sandbox/winsrt.test.ts` (Windows, both arches in CI): the wrap-shape row now asserts argv has **no** proxy flags and the returned env carries the full set (standard trio + DOCKER_HTTP_PROXY/GRPC_PROXY) at in-range ports; all existing B/C/D/E egress, bypass-resistance, relay and tool-matrix rows run against the new delivery path unchanged.
- `cargo check`, `tsc --noEmit`, `bun test` (full suite) pass locally.

## Notes

- **Twin-cased proxy variables**: the host spawn layer (Node/bun `child_process` on win32) collapses environment keys that differ only by case, keeping a single casing — so the lowercase `http_proxy`/`https_proxy`/`all_proxy` twins never survive a spawn from JS. Cygwin/MSYS2 children (case-sensitive environments, GNU tools read the lowercase names) would otherwise see no proxy and attempt direct connects that the egress sandbox blocks. The broker therefore restores the missing-case variants of `*_PROXY` variables (suffix rule — covers the standard set plus GRPC_PROXY etc.) — values remain entirely caller-specified; existing keys are never overwritten and nothing is added when none is present.
- **Cargo + revocation checking**: schannel certificate-revocation lookups go through CryptoAPI/WinHTTP, which ignores proxy env vars; the cargo test row sets `CARGO_HTTP_CHECK_REVOKE=false` (standard practice behind corporate proxies). Worth a mention in the eventual Windows README/known-limitations.
- **No broker-PID env var**: the self-protect smoke probe discovers the broker by walking the child's parent-process chain (by name, hop-capped) instead of reading a synthesized `SANDBOX_RUNTIME_WIN_BROKER_PID` variable, keeping the child environment free of srt-win-specific additions.
- **TMPDIR**: `generateProxyEnvVars` emits a POSIX `TMPDIR=/tmp/claude` for the macOS/Linux FS sandbox; the Windows wrap filters it out of the generated set (meaningless on Windows, breaks msys2 tools).


vendor/srt-win/src/launch.rs
```diff
@@ -328,63 +322,48 @@ pub fn run(spec: &ExecSpec<'_>) -> Result<u32> {
...
 
     // Surface the broker PID so the test suite can verify
     // self-protect (child tries `OpenProcess(<broker>)` → denied).
-    let pid = std::process::id().to_string();
-    env.insert(
-        "SANDBOX_RUNTIME_WIN_BROKER_PID".into(),
-        ("SANDBOX_RUNTIME_WIN_BROKER_PID".into(), pid),
-    );
+    entries.push((
+        "SANDBOX_RUNTIME_WIN_BROKER_PID".to_string(),
```
ig-ant: Could we avoid this through the smoke test probe using something like this:

```
function Get-ProcessChain($id = $PID) {
  while ($id) {
    $p = Get-CimInstance Win32_Process -Filter "ProcessId=$id"
    if (-not $p) { break }
    $p | Select-Object ProcessId, ParentProcessId, Name, CommandLine
    $id = $p.ParentProcessId
  }
}
```

and figuring out the broker process that way? It should be a fixed number of jumps from the current process.

ig-ant: Done in 7ea8873 — dropped the env var; the probe now walks the parent chain (by process name with a hop cap, since the depth differs between the cmd/powershell/git-bash rows; Win32_Process.ParentProcessId is readable without any handle on the broker, so discovery doesn't depend on the access self-protect denies). build_env_block is now a verbatim passthrough apart from restoring the twin casing of the standard proxy variables that the host spawn layer collapses (9008d9e).

vendor/srt-win/src/launch.rs
```diff
@@ -394,6 +371,47 @@ fn build_env_block(
...
+/// environment keys that differ only by case, keeping a single casing —
+/// but Cygwin/MSYS2 children import the block into a CASE-SENSITIVE
+/// environment where GNU tools read only the lowercase names. When
+/// exactly one casing of a standard proxy variable is present, append
+/// the missing twin with the identical value so the child sees the
+/// environment the caller specified. This is casing repair of
+/// caller-provided values, NOT a proxy fallback: when neither casing
+/// is present nothing is added, and no value is ever invented or
+/// rewritten. When both casings are already present nothing is added
+/// either, so no duplicate of an existing key is ever created.
```
ig-ant: If environment variables are case insensitive but GNU tools only read the lowercase names - can we get the TS side to only set the lowercase names, drop the uppercase duplicates and we can avoid this work?

ig-ant: Reworded in c87693b.

vendor/srt-win/src/launch.rs
```diff
@@ -694,29 +695,117 @@ mod tests {
...
+        assert_eq!(matching("FOO_PROXYX").len(), 1);
+        assert!(matching("foo_proxyx").is_empty());
+        assert_eq!(matching("PATH").len(), 1);
+        assert!(matching("path").is_empty());
     }
 
     #[test]
-    fn env_block_no_proxy_when_unset() {
-        let block = build_env_block(None, None);
+    fn env_block_is_verbatim_passthrough_no_proxy_synthesis() {
```
ig-ant: I think this test is of questionable value.

vendor/srt-win/src/launch.rs
```diff
@@ -694,29 +695,117 @@ mod tests {
...
-            s.split('\0').filter(|e| !e.is_empty()).collect();
-        let has = |needle: &str| entries.contains(&needle);
-        assert!(has("HTTP_PROXY=http://127.0.0.1:3128"));
-        assert!(has("https_proxy=http://127.0.0.1:3128"));
-        assert!(has("ALL_PROXY=socks5h://127.0.0.1:1080"));
-        assert!(has("NO_PROXY="));
-        assert!(has("no_proxy="));
-        // Block must end with a double-NUL.
-        assert!(block.ends_with(&[0u16, 0u16]));
+    fn proxy_case_twins_added_only_when_twin_absent() {
```
ig-ant: I think this test is of questionable value.

ig-ant: Agreed — removed in the rebase; kept only the suffix-rule test.

vendor/srt-win/src/launch.rs
ig-ant: Agreed — removed in the rebase; kept only the suffix-rule test.

dylan-conway (approved):

PR #288, opened 2026-05-29
feat(windows): srt-win acl — disk-is-truth file deny + parent allow-list with hash-ACE seal

ig-ant:
Adds `srt-win acl` for file-level filesystem deny on Windows: stamp listed files with a broker-only DACL and their immediate parent directory with a `Modify-minus-FILE_DELETE_CHILD` allow-list, so the sandboxed child cannot read/write the file's contents *or* delete/rename it via the parent. The original DACL is captured and restored on release; the **integrity of that restore record is anchored in the on-disk DACL itself** via a hash-ACE the child cannot forge.

## Invariant

> A state-DB row is a **restore record + enumeration index only** — it never asserts on-disk state. A path counts as protected iff `ensure_stamped` has, in this process under the init-mutex, read the live DACL and converged it to the broker allow-list.

There is one verification policy, enforced by type: row-level DB primitives are module-private and the only public constructor of `StampWitness` is `Locked::ensure_stamped`. A future contributor cannot add a code path that trusts a DB row without the disk check.

### Two stamps + one seal per protected file
- **File DACL** (`build_broker_only_dacl`): `PROTECTED`; group/SYSTEM/Admins `FILE_ALL`; `OWNER_RIGHTS:READ_CONTROL` (suppresses owner-implicit `WRITE_DAC`); `WriteDeny` additionally grants Everyone `GENERIC_READ|EXECUTE`.
- **Parent allow-list** (`build_parent_allow_list_dacl`): `PROTECTED (OI)(CI)`; SYSTEM/Admins/group `FA`; user `0x1301bf` (`Modify` without `FILE_DELETE_CHILD`); `OWNER_RIGHTS:READ_CONTROL`. Child cannot delete/rename the protected file but can still delete its own siblings.
- **Hash-ACE marker**: appended to *both* DACLs as `ACCESS_DENIED_ACE{sid: S-1-0-h0..h7, mask: READ_CONTROL, flags: 0}` where `h = SHA-256(original_sd || file_id)`. Inert (Null-authority SID is in no token); atomic (part of the single `SetNamedSecurityInfoW`); unforgeable by the child (no `WRITE_DAC` on either object). `mask=READ_CONTROL` not 0 because `SetNamedSecurityInfoW` and .NET `CommonAcl` silently drop mask-0 ACEs.

### `classify_sd` — the disk oracle
Reads the live DACL and returns exactly one of:
- `File(mask, marker_hash)` / `ParentAllowList(marker_hash)` — `PROTECTED` + exact broker ACE-set match (order-insensitive) + exactly one `S-1-0-*` deny-`READ_CONTROL` ACE.
- `StampedUnrecognized` — broker ACE-set match but **no marker** (marker stripped by an admin tool) **OR** marker present but ACE-set drifted (admin added an ACE). Fail-closed: never re-captured as `original_sd`.
- `Unstamped` — anything else (incl. unprotected DACLs whose inherited ACEs happen to match the broker shape — the `SE_DACL_PROTECTED` gate prevents that false positive).

### `ensure_stamped` — the chokepoint
Straight-line, no row-dependent branching on whether to verify: capture SD/file_id/links → `classify_sd` → consult DB row only as `original_sd` hint (corroborated iff `file_id` matches AND `sha256(row.original_sd || row.file_id) == on-disk marker`; tampered → FAIL; row absent → `original_sd_lost`, proceed) → record-first upsert (fence bit defaults `1`) → converge file → converge parent → derive fence bit fresh → `add_holder` → return sealed `StampWitness`.

### Restore — hash-verified, DACL-only
For each row: capture + classify. `Unstamped` and byte-equal to `row.original_sd` → already restored, drop row. `StampedUnrecognized` → `stamped_unrecognized`, leave stamped. `File(_,h)` → if `sha256(row.original_sd || row.file_id) != h` → `original_sd_tampered`, leave stamped; else `restore_sd` (writes `DACL_SECURITY_INFORMATION` only — owner/group never touched). Path-not-found / `file_id` mismatch → `locate_by_file_id(access=0)` → `relocated{movedTo}` or `missing`, row kept, stamp left in place (fail-closed). `acl recover --force` overrides the fail-closed states with a loud warning.

## State (`state_db.rs`)
- `rusqlite` (bundled, WAL, `busy_timeout=5000`) at `%LOCALAPPDATA%\sandbox-runtime\state.db`; directory carries a broker-only `(OI)(CI)` DACL.
- `acl_snapshots(canon PK, original_sd?, file_id, parent_path, parent_stamp_failed DEFAULT 1)` + `parent_stamps(parent PK, original_sd?, file_id)` + `holders(canon, pid)` + `brokers(pid, create_time)`. No `stamped_sd`/`mask` — both derivable via `classify_sd`.
- Mutex `Local\sandbox-runtime-acl-init` (broker-only DACL) wraps `acl stamp/restore/recover`. `crash_recovery` (prune dead brokers + restore orphans) runs at the top of every locked op. `WAIT_ABANDONED` is logged; recovery handles the rest.
- `exec` reads the fence plan via one `open_db_ro` connection inside one read transaction (`fence_plan_for_holder`), so the file-fence and dir-fence lists are derived from a single consistent WAL snapshot.

## Handle-fence fallback (`fence.rs`)
When the parent directory cannot be stamped (no `WRITE_DAC`) or the file has multiple hardlinks, `parent_stamp_failed=1` and `srt-win exec --holder-pid` opens only those paths with no `FILE_SHARE_DELETE` and holds them while the child runs. Stamped parent dirs are also dir-fenced (one no-share-delete handle each) to close the grandparent-rename residual.

## CLI
`srt-win acl stamp | restore | recover | status`, `--holder-pid`, `--json`, `--force`. Globs and directories rejected via a typed `CanonError`; nonexistent paths are logged as `srt-win: skipped:` and `acl stamp` exits 2 (partial). `acl restore --all --json` emits `{paths:[{path, status, expectedFileId?, movedTo?, leftStamped?, error?}], parents:[…]}`.

## Test plan
`vendor/srt-win/ci/smoke-acl.ps1` on both Windows arches:
- **A1-A25**: file stamp, refcount, crash recovery, restore round-trip fidelity, state-DB protection, mask escalation, deleted/relocated/substituted-file handling, parent allow-list (child denied del/rename, can delete sibling), `OWNER_RIGHTS` guard, ancestor `rmdir /s`, junction delete, per-exec dir fence.
- **H1** marker reaches disk with `mask=READ_CONTROL`, `flags=0`, `sha256(orig||file_id)` round-trips.
- **H2** parent marker non-inheriting. **H3** marker stable across mask escalation. **H4** poisoned `original_sd` → `originalSdTampered` fail-closed. **H5** `StampedUnrecognized` fail-closed (+ `--force` override). **H6** uncorroborated row → FAIL. **H7** wiped DB → `original_sd_lost`, marker preserved. **H8** crash-window row → next stamp re-derives. **H9** restore is DACL-only (owner untouched).
- **R2** poisoned DB row does not bypass disk-first stamp. **R820** nonexistent path skipped → exit 2; rest of batch stamped.

`cargo test`: 50 unit tests incl. `marker_ace_round_trip`, `classify_sd_calibration` (broker shapes, `StampedUnrecognized`, `SE_DACL_PROTECTED` gate), fence retry/skip, `FileId` round-trip.

## Caveats
- A relocated/tampered file's stamp is intentionally left in place (fail-closed). `--json` tells the caller; cleanup is `acl recover --force` or manual `icacls /reset`.
- `Local\` mutex is per-TS-session; same-user brokers in different terminal-server sessions share the state DB but do not mutually exclude. WAL + `busy_timeout` absorbs page-level contention; the disk-first design means a torn read never trusts a row over disk.
- Path integrity above the immediate parent (child renaming an unstamped ancestor) is a documented residual, narrowed by the per-exec dir fence on stamped parents and the state-DB dir.
- A stamp written under a different group SID classifies as `StampedUnrecognized` (fail-closed, recoverable via `--force`); cross-group recognition is a follow-up.


vendor/srt-win/src/acl.rs
```diff
@@ -0,0 +1,737 @@
...
+    if h == INVALID_HANDLE_VALUE {
+        bail!("CreateFileW('{path}'): INVALID_HANDLE_VALUE");
+    }
+    // Guard the handle so an error below doesn't leak it.
+    struct H(HANDLE);
+    impl Drop for H {
+        fn drop(&mut self) {
+            unsafe { let _ = CloseHandle(self.0); }
+        }
+    }
```
ig-ant: I think elsewhere we've had top-level types for specific handles to implement Drop on them, rather than inline Drop within a function like this. Could we do that refactor here?

vendor/srt-win/src/acl.rs
ig-ant, line 1: I think we can convert from [u8] to a concrete type and vice-versa at the boundary, and hang some functions off that type to make it easier to read the code that's doing bitwise operations.

vendor/srt-win/src/acl.rs
ig-ant: Done in 12a173b — `OwnedHandle` lives in util.rs and is used here (and in launch.rs, token.rs, and state_db.rs's process-handle paths). The other CloseHandle sites in the crate are already top-level typed wrappers (Job, InitMutex, SpawnedChild) which I left as-is.

vendor/srt-win/src/acl.rs
ig-ant: Done in 12a173b — `CapturedSd` newtype with `.control()`/`.equiv()`/`.as_bytes()`/`From<Vec<u8>>`. capture_sd/restore_sd/StampResult/Snapshot all use it; the BLOB column converts at the row-mapping boundary. The byte offsets and bit masks now live in one place behind named methods.

vendor/srt-win/src/wfp.rs
```diff
@@ -129,14 +129,16 @@ const SDDL_REVISION_1: u32 = 1;
 
 // ────────────────────── small RAII helpers ──────────────────────
 
-/// Heap SD owned by us; freed via `LocalFree`.
-struct OwnedSd {
+/// Heap SD owned by us; freed via `LocalFree`. Exposed for
+/// `state_db.rs` (init-mutex SECURITY_ATTRIBUTES) and `acl.rs`
+/// tests.
+pub struct OwnedSd {
```
ig-ant: I think OwnedSd should move to util.rs if it's used by both wfp.rs and state_db.rs

vendor/srt-win/src/main.rs
```diff
@@ -753,21 +763,77 @@ fn run() -> anyhow::Result<()> {
...
                 target_exe: &exe,
                 target_args: args,
             };
             let code = launch::run(&spec)?;
-            // Propagate the child's exit code verbatim.
+            // `_delete_fence` drops here → handles closed → fence
+            // lifted. process::exit skips destructors, so explicitly
+            // drop anything that needs cleanup BEFORE it. Propagate
+            // the child's exit code verbatim.
+            drop(_delete_fence);
```
ig-ant: Is this necessary, or will the process exiting cause the handles to be closed automatically?

vendor/srt-win/src/wfp.rs
ig-ant: Done in 6bd40fb — `OwnedSd` moved to util.rs alongside `OwnedHandle`; wfp.rs and state_db.rs both use it from there. wfp.rs keeps a local `sd_byte_blob()` for the `FWP_BYTE_BLOB` view so util stays subsystem-agnostic.

vendor/srt-win/src/main.rs
ig-ant: Not strictly — the OS closes all process handles on exit regardless of `process::exit` skipping Rust destructors. It's there as documentation-as-code: `process::exit` skips drop, so any future local that needs real cleanup (flush/sync, not just close) must be dropped explicitly above it. For `_delete_fence` specifically the OS handle reclaim is sufficient. Happy to remove the explicit drop and just keep the comment, or keep both — your call.

ig-ant: Keeping it — belt-and-braces against leaving fence handles open if anything between here and exit ever changes. Resolved.

vendor/srt-win/src/main.rs
```diff
@@ -174,6 +195,253 @@ enum GroupCmd {
...
+        group: GroupRef,
+        /// PID of the LONG-LIVED process that owns these stamps —
+        /// normally the Node host (sandbox-runtime), which calls
+        /// `acl stamp` at initialize() and `acl restore` at reset()
+        /// from a SEPARATE short-lived `srt-win` process. The stamp
+        /// persists until this PID exits or restores. Defaults to
+        /// the current (CLI) PID, which is only correct when the
+        /// same process both stamps and restores.
+        #[arg(long)]
+        holder_pid: Option<u32>,
```
ig-ant: Should pid be a type so we can avoid confusing it with other u32?

Why is this an Option<u32>? When would we stamp without providing a holder PID?

vendor/srt-win/src/main.rs
```diff
@@ -493,25 +761,354 @@ fn run() -> anyhow::Result<()> {
...
+                (&input.deny_write, acl::AclMask::WriteDeny),
+            ] {
+                for p in list {
+                    let (canon, is_dir) = acl::canonicalize_path(p)?;
+                    if is_dir {
+                        return Err(anyhow!(
+                            "Windows fs deny requires explicit file paths in \
+                             this version; got directory '{p}' (canonical \
+                             '{canon}'). Directory support is a \
+                             follow-up."
```
ig-ant: I don't think we should mention a follow-up in error messages (assuming these make it to the end user). 

Just "Windows fs deny requires explicit file paths; got directory '{p}' (canonical '{canon}')."

ig-ant: Done in b02d0a5 — now `"Windows fs deny requires explicit file paths; got directory '{p}' (canonical '{canon}')."` Also dropped the "for now" from the `Stamp` doc comment and the `acl.rs` module doc.

vendor/srt-win/src/main.rs
ig-ant: Both done in b02d0a5:
- `--holder-pid` is now **required** on `acl stamp` and `acl restore` — there's no valid use without it, since the `srt-win acl` process exits immediately and keying on its own PID would orphan the stamp instantly. (`acl recover`/`status` are holder-agnostic and don't take it; `exec`'s stays optional since exec-without-fence is valid.)
- Added `pub struct HolderPid(pub u32)` in `state_db.rs` and threaded it through `with_init_lock` / `Locked` / `fence_fallback_paths` / `query_holder_paths`. The clap arg parses as `u32` (so the struct compiles cross-platform) and is wrapped at the use site inside the `cfg(windows)` arm; the SQLite column stays `u32` with `.0` at the bind boundary.

dylan-conway (approved):

PR #296, opened 2026-06-03
fix: header bypass on tls terminate proxy

r-ushil-gsa:


shawnm-anthropic (approved):
Thank you for this fix!

PR #340, opened 2026-06-25
feat(windows): per-exec file-deny via srt-win exec --deny-read/--deny-write

ig-ant:
## What

`srt-win exec` gains repeatable `--deny-read <PATH>` / `--deny-write <PATH>` flags. When set, the exec process stamps each path broker-only under its **own PID** (via the same disk-first `ensure_stamped` chokepoint as `acl stamp`) before spawning the child, opens the same parent-stamp-fail fallback fence + parent-dir fence as the session-level path, and restores after the child exits (via the same `try_restore` chokepoint as `acl restore`). No new stamp/restore mechanism — calls the existing chokepoints under a different holder.

`wrapWithSandboxArgv` on Windows now passes `customConfig.filesystem.denyRead/denyWrite` and `customConfig.credentials.files[mode=deny]` paths through as per-exec flags (previously: threw "not supported on Windows"). Same expand pipeline as the session-level set (`expandWindowsFsDenyPaths` — normalize → glob-expand → drop-missing → reject-dir). Paths the session already covers are filtered out — re-stamping under a second holder is refcount-safe but wasteful.

**Stamp-error semantics:** any per-exec stamp failure (dir, glob, canon-fail, classify, tampered) **fails the exec** — consistent with the session-level `acl stamp` exit≠0, which `stampWindowsAcl` treats as a setup failure (throws). The exec never runs the child with an incomplete deny set. Restore is best-effort: a leftover stamp stays broker-only (fail-closed) and crash-recovery reaps it once the exec PID is dead.

## Why

Closes the per-exec deny asymmetry vs macOS/Linux: on those platforms `wrapWithSandbox(cmd, customConfig)` honours per-command `filesystem.denyRead/denyWrite` and `credentials.files[mode=deny]`; on Windows the file-deny half was rejected because the session-level stamp is applied at `initialize()`. This adds the per-exec mechanism so callers get the same per-command surface on all three platforms.

## Threat model

Same Tier-0/1 best-effort caveat as the session-level FS layer — protects against accidental and in-token deliberate access; **not** a security boundary against a deliberate sandboxed process (see the README Windows section).

## Test plan

- **F11** (`winsrt.test.ts`): `wrapWithSandboxArgv(cmd, customConfig)` with `credentials.files[mode=deny]` and `filesystem.denyRead` each produce a `--deny-read` flag; the child cannot read the target; after the child exits the broker reads it again (restored); the session-level stamp on a separate file is unaffected.
- **A26** (`smoke-acl.ps1`): `srt-win exec --deny-read $f -- cmd /c type $f` → child denied; after exec returns the file carries no marker and a fresh child (no flag) reads it.

---

*Stacked on `ig/winsrt-05-ts-fs`; will re-target to `main` after that merges.*


src/sandbox/windows-sandbox-utils.ts
```diff
@@ -753,11 +789,24 @@ export function restoreWindowsAcl(
...
+  // complete. runSrtWinJson parses stdout before checking the
+  // exit code, so a non-zero exit with the per-path JSON intact
+  // (`acl restore` prints outcomes THEN errors when any path
+  // stayed stamped) still surfaces every entry to reset()'s loop
+  // instead of collapsing to `undefined`. Only spawn-fail /
+  // unparseable output throws → log and return undefined.
   try {
-    return runSrtWinJson<WindowsAclRestoreResult>(args, { timeoutMs: 60_000 })
+    const r = runSrtWinJson<WindowsAclRestoreResult>(args, {
+      timeoutMs: 60_000,
```
ig-ant, line 803: I don't know if we can have this timeout? Some commands will legitimately take more than a minute to run e.g. user runs a C++ build or something

src/sandbox/windows-sandbox-utils.ts
```diff
@@ -868,6 +919,29 @@ export function wrapCommandWithSandboxWindows(p: WindowsSandboxParams): {
...
+  // CreateProcessW's lpCommandLine is capped at 32 767 WCHARs.
+  // Node's `shell:false` spawn builds it by MSVCRT-quoting each
+  // argv element and joining with spaces; the worst-case quote
+  // overhead per element is +2 (wrapping `"`) plus backslash
+  // doubling before embedded quotes (rare in file paths). A
+  // path-count proxy can't see this — 60 long paths overflow
+  // while 200 short ones fit — so estimate the actual command
+  // line and refuse with the same guidance the old count check
+  // gave. ~30 000 leaves headroom for the quote overhead the
+  // estimate doesn't model.
```
ig-ant, line 935: I think we need to write long commands to a file and run that instead, rather than outright failing. Maybe can be deferred to a future PR.

src/sandbox/windows-sandbox-utils.ts
ig-ant: This is the timeout on `srt-win acl restore` (and there's a matching one on `acl stamp`), not on `srt-win exec` — the wrapped command itself is spawned by the caller via the `{argv, env}` we return, with no timeout from us. 60s for restoring N file DACLs (pure syscalls, no network) is generous; we can bump it or make it `paths.length * k + base` if you'd like a margin, but a long-running build won't hit this.

src/sandbox/windows-sandbox-utils.ts
ig-ant: Agree — the right fix is `srt-win exec` reading the deny set from stdin (matching `acl stamp`'s stdin-JSON pattern) or `--deny-from-file`, instead of riding on argv. Deferrable since session-level `acl stamp` is the path for broad globs and per-exec is for the small delta on top; happy to do it in this PR if you'd rather not leave the throw.

dylan-conway (approved):

PR #347, opened 2026-06-26
feat(windows): srt-win install provisions sandbox user + user-SID WFP (v2 foundation)

ig-ant:
## What

`srt-win install` now provisions a dedicated local user account `srt-sandbox` and a local group `sandbox-runtime-users` that holds it, **alongside** the existing discriminator-group + group-keyed WFP setup. Additive — the `srt-win exec` deny-only-group path, the four-filter group-keyed WFP set, and every existing test are unchanged.

| | |
|---|---|
| **`user.rs`** (new) | `NetUserAdd`/`Del`/`SetInfo`, Winlogon `SpecialAccounts\UserList` hide, `DeleteProfileW`. 32-char password from `BCryptGenRandom` (kernel CSPRNG, rejection-sampled for uniformity). `BUILTIN\Users` resolved by well-known SID for non-English Windows. Rotation re-applies `UF_DONT_EXPIRE_PASSWD`. |
| **`sam.rs`** (new) | Canonical `NetLocalGroup*` helpers (`ensure_local_group`, `delete_local_group`, `add_member` by PSID, `is_member_of`). Both `user.rs` and `wfp::ensure_group` (now a thin shim) call it — one place to maintain the SAM idempotency rules. |
| **`dpapi.rs`** (new) | `CryptProtectData`/`Unprotect` with `CRYPTPROTECT_LOCAL_MACHINE \| CRYPTPROTECT_UI_FORBIDDEN`. |
| **`install.rs`** (new) | `write_setup`: opens `state.db` (which stamps the dir DACL — see below) and writes the DPAPI-protected credential + setup-marker fields into a `setup(key,value)` table. `read_cred`/`marker_version`/`clear_setup`. No JSON files. |
| **`state_db.rs`** | New `setup` table (`CREATE TABLE IF NOT EXISTS setup(key TEXT PRIMARY KEY, value BLOB NOT NULL)`) + `setup_get`/`setup_put`/`setup_clear`. `open_db`'s state-dir DACL stamp now also includes `(D;OICI;FA;;;<sandbox-runtime-users>)` so the sandbox account can't read `state.db` (or anything else under the dir). |
| **`wfp.rs`** | Two-filter-per-layer **user-SID-keyed** set in the same sublayer: `permit-loopback-user` (weight `0x0F80`) + `block-user` (weight `0x0F40`, `ALE_USER_ID` SD = `O:LSG:LSD:(A;;CC;;;<srt-sandbox-SID>)`). `FilterTag.user_sid` distinguishes the sets so `wfp install` (group refresh) and `install_user_filters` are independent. `wfp status` reports `user_filters`/`user_sid`; `port_range` deterministically prefers the group tag. All six `W_*` weights hoisted to module-level consts; `tests::weight_invariant` const-asserts `block-user > permit-nonmember` against the production values. New `with_wfp_txn` wraps the open→begin→commit envelope (abort-guard via `Drop`) for all four install/uninstall paths. |
| **`acl.rs`** | `stamp_dir_inheriting` gains an optional leading DENY ACE (built via SDDL). New `set_path_dacl_from_sddl` helper. |
| **`main.rs`** | `srt-win user status` / `srt-win user read-cred` (read from `state.db`); `srt-win install` runs the user step (exit 14 on failure) and only short-circuits when the user, marker version, and user-SID filters are all already present; `srt-win uninstall` clears the `setup` table and removes the user/filters unless `--keep-user`. |
| **TS** | `getWindowsSandboxUserStatus()`; `WindowsWfpStatusResult.{userFilters,userSid}`; `WindowsInstallResult.user`; `uninstallWindowsSandbox({keepUser})`. |

## Why

Foundation for running the sandboxed child **as** `srt-sandbox` (via `CreateProcessWithLogonW` from a non-elevated broker → runner → `CreateRestrictedToken` → child). A separate-user child carries a different user SID and a fresh logon session, which **structurally** closes the surrogate-spawn class — `schtasks`, `PROC_THREAD_ATTRIBUTE_PARENT_PROCESS`, BITS, RunAs="Interactive User" COM — that a same-user deny-only-group token cannot: every DACL- and WFP-keyed deny holds because the child no longer shares the real user's SID. The runner is the next piece; this PR is the install half.

The user-keyed WFP weights sit **above** the group set's `permit-nonmember`. The sandbox user is not a member of the discriminator group, so `permit-nonmember` would otherwise permit it; with `block-user` above it, the user-keyed fence is load-bearing immediately even while both sets coexist, and once the group set is removed this pair stands alone.

## Known caveats

- **Machine-scope DPAPI is not a confidentiality boundary.** Any local account — including `srt-sandbox` — can `CryptUnprotectData` a `CRYPTPROTECT_LOCAL_MACHINE` blob if it can read it. The credential lives in `state.db`; the state directory carries an explicit `(D;OICI;FA;;;<sandbox-runtime-users>)` DENY plus the broker-only `PROTECTED` allow set, and `state_db::open_db` re-stamps that shape on every `acl stamp`. **The directory DACL is the only gate.**
- **Idempotent same-config re-install does not rotate the password** — the "already installed, no changes" early-return checks the WFP set, the user, the marker version, and the user-SID filters, and fires before the user step. Pass `--force` to rotate.
- Three pre-existing `@typescript-eslint/await-thenable` lint errors in `winsrt.test.ts` (lines 374/525/1352) are inherited from `main` (#340); not introduced here.

## Test plan

- `cargo test`: DPAPI round-trip, password shape/uniqueness, `BUILTIN\Users` name resolution, SDDL templates, `FilterTag` user-set round-trip, weight invariant `block-user > permit-nonmember`.
- `smoke.ps1`: post-install `user status` from `state.db` (exists, SID, group memberships, hidden, `cred_present:true`, `marker_version:1`); state-dir DENY ACE present; `user read-cred` → 32-char password; `wfp status` → ≥4 user-SID filters tagged with the provisioned SID; real-user `curl https://example.com` → 200 (`block-user` doesn't match the real user); `wfp install` group-refresh leaves user filters intact; `uninstall --keep-user` keeps user + setup rows; plain `uninstall` removes user + setup rows + filters.
- `winsrt.test.ts`: `installWindowsSandbox` round-trip asserts `user.{provisioned,sid,…}` + `wfp.{userFilters,userSid}` and post-uninstall absence. Existing B/C/D/E/F rows unchanged.
- The sandbox-side "is `srt-sandbox` blocked" and "can `srt-sandbox` read `state.db`" assertions need the runner to launch a process as that user — deferred to that PR's smoke.


vendor/srt-win-src/src/install.rs
```diff
@@ -0,0 +1,132 @@
...
+    state_db::setup_put(&conn, KEY_CRED, &cred)?;
+    state_db::setup_put(
+        &conn, KEY_MARKER_VERSION, &SETUP_VERSION.to_le_bytes(),
+    )?;
+    state_db::setup_put(&conn, KEY_SANDBOX_USER, u.username.as_bytes())?;
+    state_db::setup_put(&conn, KEY_SANDBOX_USER_SID, u.sid.as_bytes())?;
+    state_db::setup_put(
+        &conn, KEY_SANDBOX_GROUP_SID, u.group_sid.as_bytes(),
+    )?;
+    state_db::setup_put(&conn, KEY_CREATED_AT_UNIX, &now.to_le_bytes())?;
```
ig-ant: I think state_db should take a struct with this info, and it can perform the inserts behind the scenes - instead of having to leak the setup_put everywhere which is a very thin wrapper over INSERT.

vendor/srt-win-src/src/install.rs
```diff
@@ -0,0 +1,132 @@
...
+            SETUP_VERSION
+        ));
+    }
+    let user = state_db::setup_get(&conn, KEY_SANDBOX_USER)?
+        .map(String::from_utf8)
+        .transpose()
+        .context("sandbox_user is not UTF-8")?
+        .unwrap_or_else(|| user::SANDBOX_USER.into());
+    let pw = String::from_utf8(dpapi::unprotect(&blob)?)
+        .context("password is not UTF-8")?;
```
ig-ant, line 84: Again I think state_db should have a way to get a user, which'll do the required db commands and hydrate a struct.

vendor/srt-win-src/src/user.rs
```diff
@@ -0,0 +1,482 @@
...
+//! ## Why a group AND a user
+//!
+//! [`SANDBOX_GROUP`] is the trustee for the credential-file DENY ACE
+//! and (later) the working-tree grant. Keying those on the *group*
+//! rather than the user SID means a future multi-account design
+//! (e.g. per-session sandbox users) only adds members; the DACLs
+//! and WFP filters don't change. It is **distinct** from the
+//! discriminator group (`sandbox-runtime-net`) — the real user is a
+//! member of *that* one, the sandbox user is a member of *this* one,
+//! and never both.
```
ig-ant, line 27: I don't think we should have the real user be part of a discriminator group long term; I don't think it's necessary and it complicates setup.

ig-ant: Agree — that's the v2-e endpoint: drop `sandbox-runtime-net` entirely; the real user is in no special group; state-db dir ACL keys on `<real-user-SID>` (with install passing the pre-elevation user SID through `maybe_self_elevate` to handle over-the-shoulder UAC); WFP is the 2-filter user-SID set only; file/parent stamps key on `<real-user-SID>` (v2-c). This PR keeps the discriminator group only because the v1 path is still load-bearing in the additive phase (v1's F0-F3 WFP, file stamps, and the state-db dir DACL all key on it today). v2-e removes it outright rather than gating it.

vendor/srt-win-src/src/install.rs
ig-ant: Done — added `state_db::SetupInfo` plus `write_setup_info` / `read_setup_info` / `clear_setup_info`; `setup_get`/`setup_put` are now module-private and the key constants moved into `state_db`. `write_setup` builds a `SetupInfo` and hands it to `write_setup_info` (which writes all six rows in one tx). In 66b6505.

vendor/srt-win-src/src/install.rs
ig-ant: Done — `state_db::read_setup_info(&conn) -> Result<Option<SetupInfo>>` hydrates the whole record (`None` when `marker_version`/`cred` absent). `install::read_setup()` is the no-conn convenience wrapper; `read_cred` and `main.rs` (`user status` + the install early-return) now go through it instead of poking individual rows. `marker_version`/`marker_version_on` are gone. In 66b6505.

vendor/srt-win-src/src/state_db.rs
```diff
@@ -135,6 +135,12 @@ CREATE TABLE IF NOT EXISTS parent_stamps (
 CREATE INDEX IF NOT EXISTS holders_by_pid ON holders (pid);
 CREATE INDEX IF NOT EXISTS snapshots_by_parent
   ON acl_snapshots (parent_path);
+-- Install-time key/value: the sandbox-user credential (DPAPI blob)
+-- and the setup marker. Additive table — no schema-version bump.
+CREATE TABLE IF NOT EXISTS setup (
+  key   TEXT PRIMARY KEY,
+  value BLOB NOT NULL
+);
```
ig-ant, line 149: I think we should have a proper table for users, instead of a key / value pair. This smells like a lift-and-shift from the JSON approach.

ig-ant: Replaced the kv `setup` table with a typed-column `sandbox_user` table (`username` PK, `user_sid`, `group_sid`, `cred` BLOB, `marker_version`, `created_at_unix`). `write_setup_info` is now a single `INSERT OR REPLACE`, `read_setup_info` a single `SELECT … LIMIT 1`; `setup_get`/`setup_put` and the `KEY_*` consts are gone. Kept the struct as `SetupInfo` (it's the install-time setup record; field names unchanged so install.rs/main.rs are untouched). In 1c0adf9.

dylan-conway (approved):

PR #377, opened 2026-07-03
proxy: flush tunneled bytes before closing the client socket

ehsan:
The CONNECT and SOCKS relays tore the client socket down the moment the upstream socket closed:

    upstream.pipe(socket)
    socket.pipe(upstream)
    socket.on('close', () => upstream.destroy())
    upstream.on('close', () => socket.destroy())   // <-- data loss

An HTTP/1.1 server that honors `Connection: close` writes its response and immediately sends FIN. The upstream socket fully closes as soon as that FIN is read (net.Socket defaults to allowHalfOpen: false), so the 'close' handler runs while relayed response bytes are still queued in the client socket's write buffer -- and destroy() discards them. The client sees a clean EOF mid-body: silent truncation, no error at either end. How much is lost depends on how much of the response is still in flight when the FIN is processed.

Keep-alive clients never trigger it, so most tools work through the sandbox. Clients that send `Connection: close` on every request hit it reliably for any response larger than the socket buffers can absorb. Python's stdlib is the notable one -- urllib hardcodes that header, so inside the sandbox:

    $ python3 -c "import urllib.request as u; \
        u.urlopen('https://pypi.org/simple/pip/').read()"
    http.client.IncompleteRead: IncompleteRead(81920 bytes read,
                                               23693 more expected)

while the same fetch works outside the sandbox, with keep-alive clients (curl, requests, pip), and for small responses -- which makes the failure look like anything but the proxy.

Fix: on upstream 'close', end() the client socket instead of destroying it. pipe() already propagates the upstream's graceful EOF as socket.end(), which flushes the write queue before sending FIN; the explicit end() is a no-op in that path (and after the error path's destroy), and a flush-then-FIN for the remaining cases. Teardown still converges: the client's own close destroys the upstream, and upstream 'error' still destroys the client socket immediately.

Both relays got the destroy-on-close teardown with the parent-proxy support in fd74a3f (#187); the relay before that propagated 'end' instead.

The regression test tunnels 1 MiB close-after-write streams through both proxies to a deliberately slow reader connected over a Unix domain socket -- its small fixed kernel buffers keep relayed bytes queued in userspace exactly when the upstream FIN lands, so the loss reproduces reliably (fails 5/5 suite runs before the fix, passes 10/10 after; over loopback TCP the kernel absorbs a few hundred KB and hides the race most of the time).


CroneKorkN:
Independent reproduction of this bug, arrived at from the symptom side (empty
replies through the Claude Code sandbox proxy) and traced back to the same two
lines. Posting the evidence here rather than opening a duplicate issue; the
diagnosis matches this PR's, and the measurements add a macOS/Bun angle plus a
second, non-code confirmation of the mechanism.

## What we saw

Total loss, not truncation: 0 response bytes after a successful `CONNECT` 200,
TLS handshake and request. Python `http.client` reports `RemoteDisconnected`,
curl `(52) Empty reply from server`. Target was a private HTTPS origin
returning a 7,728-byte JSON body and closing the connection afterwards.

Trigger is `Connection: close` on the request, not the client implementation
(urllib hardcodes the header, curl never sends it by default). Interleaved
12x6 matrix inside the sandbox, Claude Code 2.1.274, macOS arm64:

| client arm | header | failed |
|---|---|---|
| `urllib.request` | `Connection: close` (implicit) | 4/12 |
| `http.client`, explicit | `Connection: close` | 5/12 |
| raw CONNECT + ssl + GET | `Connection: close` | 2/12 |
| `curl --http1.1 -H 'Connection: close'` | `Connection: close` | 6/12 |
| `http.client`, keep-alive | - | 0/12 |
| `curl --http1.1` | - | 0/12 |

## Why macOS/Bun makes it loud, and why small responses are the worst case

Claude Code bundles srt into a Bun binary, and its HTTP path is
client -> mux (TCP) -> `http-proxy` over a **unix socket** -> upstream. On
macOS `net.local.stream.sendspace`/`recvspace` default to **8192**, so a
7.7 KiB response does not fit the AF_UNIX buffer in one write and the
remainder stays in the stream's writable queue. Bun then emits `'close'` on
the upstream socket with no event-loop turn between `'end'` and `'close'`, so
`socket.destroy()` runs before anything can drain. Node emits `'close'` one
loop iteration later, which usually lets the queue flush - hence the very
different rates (roughly 25-50 % on Bun, ~5 % on Node, same origin, same
run).

Note for the PR description: the loss is not limited to "responses too big for
the socket buffers" in the large sense. In our size sweep it is *small*
responses that fail hardest (~5.6 KiB: 10/24; ~115 KiB: 0/12; ~712 KiB: 1/12),
because what matters is whether the *last* chunk before the FIN is still
queued, not the total size. A big transfer usually has an empty queue at FIN
time; a small one arrives as a single burst together with close_notify and
FIN.

## Per-request correlation

Instrumented dist (npm 0.0.76), Bun 1.4.2, mux on, `Connection: close` arm,
16 requests, 6 failures. Measured at the moment the `'close'` handler calls
`socket.destroy()`:

| outcome | `socket.writableLength` in http-proxy | mux `client.writableLength` |
|---|---|---|
| 6x FAIL | 3531, 3531, 4730, 4730, 7128, 8327 (`writableFinished=false`) | 0 |
| 10x OK | 0 (once 24 = TLS close_notify) | 0 |

The correlation is complete: every failure has an unflushed writable queue on
the unix-socket hop, every success has none. The mux hop was clean in all 16
cases, which is why the effect is entirely in `http-proxy`'s teardown - the
line this PR changes.

## A/B against the fix

We patched the npm 0.0.76 dist with `destroySoon()` (functionally the same as
this PR's `end()` for this path: `destroySoon` is `end()` plus a destroy once
`'finish'` fires) and ran interleaved A/B, same runtime and origin, mux on:

| runtime | variant | `Connection: close` | curl + `Connection: close` | keep-alive |
|---|---|---|---|---|
| Bun 1.4.2 | 0.0.76 dist | 9/24 | 6/24 | 0/24 |
| Bun 1.4.2 | destroySoon | 0/24 | 0/24 | 0/24 |
| Node 26.0.0 | 0.0.76 dist | 1/12 | 0/12 | 0/12 |
| Node 26.0.0 | destroySoon | 0/12 | 0/12 | 0/12 |

Bun: 15/48 -> 0/48 on the two affected arms (0/72 including keep-alive).
Node: 1/36 -> 0/36.

## Independent confirmation of the mechanism, without touching the code

Raising the AF_UNIX buffer so the last chunk fits removes the loss entirely,
which is what the queue-discard explanation predicts:

- `sudo sysctl -w net.local.stream.sendspace=262144 net.local.stream.recvspace=262144`
  **before** starting the process: 0/144 failures across two 12x6 matrices
  against the real (unpatched) Claude Code proxy - the same arms that were at
  17-50 % on the 8192 default.
- The same sysctl applied to an **already running** proxy: no effect
  (5/12, 5/12, 2/12, 2/12 on the four `Connection: close` arms), because
  `sendspace`/`recvspace` only set the default `SO_SNDBUF` at socket creation
  time and the proxy binds its mux listener and internal unix socket once per
  session.
- Standalone A/B with the 0.0.76 dist, same picture: ~5.6 KiB response
  10/24 + 10/24 at 8192, 0/24 + 0/24 at 262144.

That is a symptom shifter, not a fix (it moves the threshold from 8 KiB to
256 KiB and is system-wide), but it is a second, code-independent line of
evidence for the writable-queue diagnosis.

## State of `main` and a suggestion

On `main` (1831bab, 2026-09-16, package.json 0.0.76) the handlers are
unchanged and have moved to `src/sandbox/http-proxy.ts:613-614`:

```ts
socket.on('close', () => upstream.destroy())
upstream.on('close', () => socket.destroy())
```

so this PR likely needs a rebase (it was written against ~line 281).

Two suggestions, offered as observations rather than requests:

1. The same idiom sits in `src/sandbox/mux-proxy.ts:134-135` and is not
   covered by this PR:

   ```ts
   client.once('close', () => upstream.destroy())
   upstream.once('close', () => client.destroy())
   ```

   In our 16-request correlation the mux hop was always clean, so we have no
   failure to attribute to it - but it is the same teardown-before-flush
   pattern on a second relay, and in the Claude Code topology every HTTP byte
   passes through it.

2. Applying the flush to *both* directions (the `socket.on('close')` handler
   too) covers the mirror case, a client that closes while request bytes are
   still queued toward the upstream. Untested here; symmetric for the same
   reason.

Incidentally, `http-proxy.ts:231-243` already treats this exact race class on
the status-write path ("Destroy only after the flush: destroying immediately
races the ..." with `end(..., cb -> destroy)` plus a backstop timer), which
suggests the relay teardown is simply the site that was missed.

Downstream user report of the same failure: anthropics/claude-code#81136.

(ai generated)

PR #395, opened 2026-07-10
feat(windows): async dependency-check APIs + configurable install timeout

ig-ant:
Two problems with the Windows dependency/status/install helpers:

1. **They are `spawnSync`-backed.** Every probe freezes the embedder's event loop for ~70–85 ms warm — seconds when Defender cold-scans `srt-win.exe` — forcing embedders to route what should be cheap status checks off their hot/render paths.
2. **`installWindowsSandbox` hard-kills at 60 s while the UAC consent dialog lives ~2 min.** A slower legitimate approval gets killed mid-elevation, and elevation is not retracted when the wrapper dies — a late click half-completes the install.

What this PR does:

- Async variants end-to-end: `runSrtWinAsync`, `getWindowsWfpStatusAsync`, `getWindowsSandboxUserStatusAsync`, `installWindowsSandboxAsync` (readback probes run concurrently), `checkWindowsDependenciesAsync` (both probes concurrently). Argv-construction and JSON-mapping are hoisted into shared helpers so sync and async are byte-identical for the same underlying state; the sync originals stay exported.
- `SandboxManager` adds `checkDependenciesAsync()`; the sync `checkDependencies()` is unchanged (Windows callers should prefer the async variant to avoid event-loop stalls; on POSIX the checks are native and async simply wraps the sync result). `initialize()` uses the async helpers internally.
- `installWindowsSandbox{,Async}` and `uninstallWindowsSandbox` gain `opts.timeoutMs`, default **120 000 ms** to match the UAC dialog TTL (uninstall was falling through to the 15 s `runSrtWin` default).

Tests: cross-platform sync/async parity suite (spawn/spawnSync spied) pinning identical argv, result mapping, and throw messages; concurrent-probe behavior; the 120 s default + override.


src/sandbox/sandbox-manager.ts
```diff
@@ -1994,7 +1973,7 @@ export interface ISandboxManager {
   checkDependencies(ripgrepConfig?: {
     command: string
     args?: string[]
-  }): SandboxDependencyCheck
+  }): Promise<SandboxDependencyCheck>
```
ig-ant, line 2048: Note for @dylan-conway - this is a public API change; is that okay?

ig-ant: Resolved; we changed it to add a fresh checkDependenciesAsync instead of modifying checkDependencies

dylan-conway (approved):

PR #462, opened 2026-08-11
feat(windows): machine-wide state store shared by all users

ig-ant:
## Summary
The Windows sandbox kept its credential and managed CA in a per-user store (`%LOCALAPPDATA%\sandbox-runtime`). That breaks:

- **Fleet installs running as SYSTEM** (SCCM/Intune): the credential lands in SYSTEM's profile, so real users' dependency checks see `cred_present:false`.
- **Multi-user machines**: each user's install rotates the shared account's password, invalidating every other user's stored copy.

This makes `%ProgramData%\sandbox-runtime` — shared by all users — the one and only store. The Windows sandbox is pre-release, so the per-user store is removed outright (no resolution fallback, no migration machinery); existing pre-release installs just re-run the elevated install.

**How**
- The elevated install provisions the dir: ownership to Administrators + PROTECTED inheritable DACL (SYSTEM/Admins full, Users modify, sandbox group DENY). Ownership matters — ProgramData lets standard users pre-create paths, and a squatter's creator-owner keeps `WRITE_DAC` through any DACL we write.
- The credential moves out of the DB into `cred.dat` with a stricter DACL (Users read-only, Admins write, sandbox DENY, ownership taken). The DB's `cred` column is gone (schema v8) — the Users-writable `state.db` never carries a credential.
- `state_dir()` (Rust) and `windowsStateDir()` (TS) are the same pure env-derived path — no fs probing.
- The install "already installed" early-out now includes `cred.dat` presence, so a deleted/tampered credential file is rewritten by re-install instead of early-outing.
- The acl init mutex moves to `Global\` (cross-user serialization) with a sandbox-deny DACL and a bounded 60s wait.

**Accepted trade-off** (decided): any real local user can read the shared credential or replace the CA — the sandbox account stays network-confined by SID-keyed WFP filters regardless of spawner, and concurrently-interactive multi-user machines are rare. Residual: a squatter on the Global mutex name can stall acl ops (bounded, loud, DoS-only — never a bypass).

## Test plan
New `ci/smoke-machine-store.ps1`, wired into the Windows CI legs and verified end-to-end on a Windows 11 arm64 guest:

- ACL/ownership assertions on the store and `cred.dat`; sandbox child denied on both `cred.dat` and `state.db`.
- **Install as SYSTEM via schtasks, then the interactive user reads the credential and spawns** — the fleet-shape regression.
- Rotation refreshes the shared copy in place; uninstall removes the credential with the account.

Cross-target `cargo clippy` zero warnings; `tsc --noEmit` clean; TS suites unchanged vs main.


vendor/srt-win-src/src/state_db.rs
```diff
@@ -630,7 +611,9 @@ pub fn with_init_lock<R>(
     force_recover: bool,
     f: impl FnOnce(&mut Locked) -> Result<R>,
 ) -> Result<(R, RecoveryReport)> {
-    let (_mutex, conn, report) = locked_recovered(force_recover)?;
+    // Unelevated broker context: recovery applies disk ACEs (the
+    // caller's own token bounds what SetNamedSecurityInfoW can do).
+    let (_mutex, conn, report) = locked_recovered(force_recover, DiskAces::Apply)?;
```
cirospaciari, line 616: Now that state.db is BUILTIN\Users-modify, doesn't the unelevated path still trust the rows? As far as I can see crash_recovery takes the SID from the raw `sandbox_user` row (not the validated `read_setup()`) and recomposes whatever `working_aces`/`ace_holders` rows it finds — so couldn't user B insert a held grant row for one of A's paths (or point `user_sid` at A's own SID) and have A's next acl op apply/strip ACEs on A's tree with A's token? Or is there a per-user filter I'm missing?

vendor/srt-win-src/src/state_db.rs
```diff
@@ -1243,37 +1229,34 @@ pub fn set_ambient_denies(
...
-        // leftover ACE stays visible to `status` and retryable by a
-        // later uninstall/install instead of becoming an invisible
-        // orphan.
-        conn.execute(
-            "DELETE FROM ambient_denies WHERE canonical_path = ?1",
-            params![p],
-        )
-        .context("DELETE ambient_denies")?;
-        match recompose_at(conn, &p, sandbox_sid) {
+    for raw in crate::ambient::ambient_deny_targets() {
```
cirospaciari, line 1243: Does the same reasoning apply to `set_ambient_denies` above? It's also under `with_install_lock` but still goes through `recompose_at`, which folds `kind`/`mask` from the (now Users-writable) `working_aces` rows into what gets applied with SeRestore — so could an inserted deny / garbage-mask row for one of the static targets get stamped (or fail every target) on the next elevated install?

vendor/srt-win-src/src/state_db.rs
```diff
@@ -1413,6 +1410,10 @@ fn crash_recovery(conn: &Connection, force: bool) -> Result<RecoveryReport> {
                 params![&canon, &kind],
             )
             .context("DELETE working_aces (orphan)")?;
+            if disk == DiskAces::Skip {
+                report.aces_revoked += 1;
```
cirospaciari, line 1414: If the row is already deleted by the time we `continue` here, what removes the on-disk ACE later? `acl recover` / the with_init_lock recovery look row-driven to me, so once an elevated install (or `uninstall --keep-user`) prunes a crashed session's rows wouldn't the srt-sandbox ALLOW stay on the path for good? Is there a SID-based sweep I'm not seeing? (Also `aces_revoked` gets bumped here without anything being revoked.)

vendor/srt-win-src/src/install.rs
```diff
@@ -26,18 +26,137 @@ use crate::{dpapi, logon, runner, user};
...
+    // child tamper with the refcount. Heal ownership on every
+    // elevated install; the inherited DACL then governs.
+    for f in ["state.db", "state.db-wal", "state.db-shm"] {
+        let p = dir.join(f);
+        if !p.exists() {
+            continue;
+        }
+        if let Some(ps) = p.to_str()
+            && let Ok(fh) = crate::acl::open_for_security_no_follow(ps)
+            && let Err(e) = crate::acl::set_handle_owner_admins(&fh, f)
```
cirospaciari, line 116: Is healing only the owner enough here? Explicit ACEs already on a pre-existing state.db would survive the parent DACL rewrite and get evaluated before the inherited sandbox DENY, so a state.db squatted with e.g. an explicit Everyone allow would still be reachable from the sandbox after install, wouldn't it? And `ca\` / a state.db re-created after the schema rename don't seem to get re-owned at all — should this rewrite (or just delete) the children instead?

cirospaciari (commented):
A few questions on the untrusted-rows side of this, not sure I have all of them right.

vendor/srt-win-src/src/state_db.rs
ig-ant: You're right — and it goes deeper than the raw-SID read: even with a validated SID, any local user could seed `working_aces` rows against another user's paths for that user's own broker to apply. There's no per-user filter; the shared Users-writable DB was the wrong home for session bookkeeping. #467 restructures: session rows move back to a per-user `%LOCALAPPDATA%` store (real-user-only DACL), so rows are only writable by the user whose broker consumes them, and recovery takes its SID from the SAM-validated machine-store read.

vendor/srt-win-src/src/state_db.rs
ig-ant: Correct — `set_ambient_denies` had the same hole: `recompose_at` folds kind/mask from Users-writable rows under an elevated token (garbage mask ⇒ every target fails; forged grant row ⇒ stamped with SeRestore). In #467 both ambient stamp and clear apply `SbAceSet` directly from the static env-derived target list and never recompose through rows; session recomposes read only the per-user store plus a deny-only ambient fold (a forged ambient row can only add a write-deny, never a grant).

vendor/srt-win-src/src/state_db.rs
ig-ant: Right on both counts: `acl recover` is row-driven (no SID sweep exists), so pruning rows without disk work orphaned the ACE permanently, and the `aces_revoked` bump was phantom. In #467 the rows-only mode is gone entirely — elevated flows never touch session DBs, and session recovery is back to full row-driven semantics in the per-user store, so nothing is pruned without its disk ACE being converged.

vendor/srt-win-src/src/install.rs
ig-ant: Agreed — owner healing alone left explicit squatter ACEs on children evaluating ahead of the inherited deny, and `ca`/post-rename DBs weren't healed. #467 keeps ownership healing for the machine store's children and additionally: a corrupt/garbage `install.db` is renamed away and recreated on the next elevated install (it also degrades to not-provisioned on the read path rather than erroring, so a planted file can't wedge every broker), and the session `state.db` no longer lives in the shared store at all.

PR #479, opened 2026-08-15
fix(seccomp): fall back to fork() when unshare(CLONE_NEWUSER) hits EINVAL

Xiaokebuyu:
## Problem

`unshare(CLONE_NEWUSER)` implies `CLONE_THREAD`, and `check_unshare_flags()` rejects that with `EINVAL` while `thread_group_empty()` is false. A thread that has already exited stays in the thread group until `release_task()` runs, but its joiner is woken earlier (in `mm_release()`), so code that calls `unshare` right after joining a thread can land inside that window. With a helper thread still running, it fails every time.

The standalone `apply-seccomp` binary cannot hit this: `execve()` leaves it single-threaded. The exposed configuration is `seccomp.argv0`, a multicall binary running this sequence in-process. anthropics/claude-code#86928 reports it from a bun-based embedder (`apply-seccomp: unshare(CLONE_NEWUSER): Invalid argument` on roughly 1 in 10 sandboxed commands), corroborated on x86_64 bare metal, rootless Docker and a multi-user login node.

## Fix

When `unshare(CLONE_NEWUSER)` fails with `EINVAL`, `fork()` and redo the unshare in the child, which is single-threaded. The parent stays behind as a relay that forwards signals and exits with the child's status, the same way the outer stub already relays the inner init. It closes its copies of the observation socketpair and the pinned `/proc` fd, which the child now owns.

The common path is unchanged except for one reorder: `unshare(CLONE_NEWUSER)` now runs before dumpable is raised rather than after. That way a relay left behind by the fallback keeps the process's original dumpable state. `unshare` itself does not change dumpable, and with a mode-0111 install every process in the chain stays non-dumpable.

## Why not `clone(CLONE_NEWUSER|…)` or a retry

- **Retry** (the first revision of this PR) only covers a thread that has exited but has not been released yet. With a live helper thread, the retry loop just runs out its budget.
- **Raw `clone(SIGCHLD|CLONE_NEWUSER|CLONE_NEWPID|CLONE_NEWNS)`** (the second revision) has no thread-group check. However, a raw clone skips the embedder's atfork handlers and glibc's own fork-time lock reset (malloc arenas, stdio; `posix/fork.c`). If a live thread holds one of those locks at that moment, the child's later `fork()` of the worker, or `perror()` on an error path, blocks forever. A clear `EINVAL` turns into a silent hang. `fork()` keeps those guarantees.

## Measurements

In-process harness: the variant's `main()` is called from a host binary after the thread setup below. Single-vCPU aarch64, Linux 6.1, glibc 2.36. A run counts as a hang if it has not exited after 1.5 s.

| thread state when `main()` runs | `main` (unshare) | raw clone (rev. 2) | this PR |
|---|---|---|---|
| helper thread just joined | 50/50 `EINVAL` | 50/50 ok | 50/50 ok |
| live thread cycling a lock taken by an atfork handler | 50/50 `EINVAL` | **50/50 hang** | 50/50 ok |
| live thread cycling `malloc`/`free` | 50/50 `EINVAL` | **21/50 hang** | 50/50 ok |

Also verified with this change:
- exit codes 42 and 143 (SIGTERM to the outermost process) are relayed, with no stray processes left behind;
- with a mode-0111 install, the relay, stub and inner init all stay non-dumpable;
- 100/100 runs nested inside an unprivileged `bwrap --unshare-user --unshare-pid`;
- `SRT_OBSERVE_SOCK` output matches the standalone binary;
- on this host, `test/sandbox/{seccomp-filter,pid-namespace-isolation,execute-only-binary,linux-violation-monitor,mandatory-deny-paths,allow-read,integration}.test.ts` give the same results as `main`. The one failure is an integration case that needs `nc`, which this host does not have.

## Test

The fallback cannot be reached through the standalone binary, so `test/sandbox/apply-seccomp-einval-fallback.test.ts` injects `EINVAL` with strace. It checks two cases:
- the fallback runs the command and relays its exit status;
- a setup failure inside the forked child still fails closed.

Both tests fail against `main`'s binary and pass with this change. The CI Linux job now installs `strace`. The suite skips when strace or ptrace is unavailable.

A note on scope: on the Claude Code build I checked (2.1.272, arm64), `ARGV0=apply-seccomp` no longer starts a thread before this code runs (strace shows no `CLONE_THREAD`), so that build does not trigger the race today. This change is a defense for any multithreaded `seccomp.argv0` embedder rather than a dependency on one embedder's startup order.


Xiaokebuyu:
Rebased onto main and reworked: the retry loop is gone. The nested user namespace is now created on the clone() that the outer stub already performs for the inner init, so there is no unshare(CLONE_NEWUSER) left to race — clone() has no thread-group check, and this also covers a helper thread that is still alive, where a retry could never succeed. Title, description and commit message updated; measurements are in the description.

Xiaokebuyu:
Reworked again: the raw `clone()` revision is replaced by a `fork()` fallback on `EINVAL`.

My previous comment said the clone approach also holds while a helper thread is still alive. That was measured with an idle thread, and it does not hold in general. A raw clone skips atfork handlers and glibc's fork-time lock reset. With a live thread cycling an atfork-guarded lock, the clone child hung in its later `fork()` 50/50 times; with a thread cycling `malloc`/`free`, it hung 21/50 times. The fork fallback completed 50/50 in both cases. Details and the updated table are in the description.

This revision also adds a strace fault-injection test for the fallback path, and adds `strace` to the CI Linux dependencies.

Issue #419, opened 2026-07-22
[BUG] macOS Seatbelt profile denies tty ioctls (TIOCSETA/TIOCSETAW), breaking interactive TUIs run under `srt`


marcoscano:
### Preflight
 
- [x] I have searched existing issues and this hasn't been reported yet
- [x] This is a single bug report (not multiple bugs)
- [x] I am using the latest version of `@anthropic-ai/sandbox-runtime`

## Summary
 
Any process running under `srt` on macOS fails to put its stdin into raw mode: `tcsetattr` returns `EPERM` because the generated Seatbelt profile denies `file-ioctl` on the controlling terminal device. This breaks every interactive TUI run under the sandbox, including Claude Code , which the [Claude Code docs](https://docs.claude.com/en/docs/claude-code/sandbox-environments) explicitly recommend launching as `npx @anthropic-ai/sandbox-runtime claude`.
 
Adding the tty device to `filesystem.allowWrite` does **not** resolve it, which suggests `allowWrite` emits `file-write-data` but not `file-ioctl` for the matched paths. I could not find any configuration option that reaches this.
 
## Environment
 
- **srt version:**  0.0.66
- **OS:** macOS Sequoia 15.7.7
- **Terminals tested:** iTerm2 and Terminal.app (identical behaviour in both)
- **Claude Code version:** 2.1.217
- **Node:** v23.10.0

## Reproduction
 
Minimal repro, no Claude Code required:
 
```bash
npx @anthropic-ai/sandbox-runtime node -e "try{process.stdin.setRawMode(true);console.log('raw ok');process.stdin.setRawMode(false)}catch(e){console.log('raw FAILED:',e.message)}"
```
 
**Actual:**
 
```
raw FAILED: setRawMode EPERM
```
 
**Expected:** `raw ok`. The same command outside the sandbox succeeds.
 
## Impact on Claude Code
 
```bash
npx @anthropic-ai/sandbox-runtime claude --dangerously-skip-permissions
```
 
Claude Code starts but never enters raw mode, so the terminal's replies to its capability queries (XTVERSION, DA1) are echoed to the display instead of being consumed from stdin, and keyboard input is line-buffered rather than handled per keystroke. The session is effectively unusable:
 
```
^[P>|iTerm2Claude1Code[v2.1.217;4;6;17;18;21;22;52c ▐▛███▜▌
▝▜█████▛▘  Opus 4.8 (1M context) with xhigh effort · Claude Max
[...]

Issue #391, opened 2026-07-09
KKP escape sequences leak as literal text through sandbox-exec PTY layer on macOS


slowmotion:
## Bug: KKP escape sequences leak as literal text through sandbox-exec PTY layer on macOS

**Environment**
- macOS (Apple Silicon / Intel)
- `@anthropic-ai/sandbox-runtime` (latest)
- Terminal emulators affected: iTerm2, Ghostty
- Command: `srt claude --dangerously-skip-permissions`

---

## Description

When running `srt claude` on macOS, terminal escape sequences are not properly consumed by the sandboxed process and instead appear as literal text in the Claude Code input prompt.

The root cause appears to be that `srt`'s `sandbox-exec` (Seatbelt) PTY layer does not correctly handle modern terminal protocols — specifically the Kitty Keyboard Protocol (KKP). Pressing Ctrl+C or Ctrl+D produces raw KKP sequences as visible text instead of being handled as control characters:

```
^[[99;5u   ← Ctrl+C  (CSI 99 ; 5 u, codepoint 'c' with Ctrl modifier)
^[[100;5u  ← Ctrl+D  (CSI 100 ; 5 u, codepoint 'd' with Ctrl modifier)
```

On iTerm2, mouse tracking sequences also leak:

```
^[[I^[[O;1R^[[<35;20;23M^[[<35;21;24M ...
```

---

## Steps to reproduce

1. macOS with iTerm2 or Ghostty installed
2. `npm install -g @anthropic-ai/sandbox-runtime`
3. `npm install -g @anthropic/claude-code`
4. `srt claude`
5. Press Ctrl+C or Ctrl+D, or move the mouse over the terminal

**Expected:** normal control character behavior (interrupt, EOF, no visible output)  
**Actual:** raw escape sequences appear as literal text in the Claude Code input prompt

---

## What was ruled out

Running `claude` directly (without `srt`) works correctly in both terminals — no sequences leak. The issue is specific to the `sandbox-exec` PTY layer.

The following workarounds were attempted without success:

- `printf '\e[?1004l\e[?1000l\e[?1002l\e[?1003l\e[?1006l' && srt claude`  
  — Claude Code re-enables mouse tracking after its TUI initializes

- `CLAUDE_CODE_DISABLE_MOUSE=1 srt claude`  
  — Fixes mouse tracking sequences on Ghostty, but KKP sequences (Ctrl+C/D)
  still leak

[...]

PR #480, opened 2026-08-16
fix(macos): grant the inherited terminals so sandboxed TUIs can enter raw mode

pcontrerasp:
Fixes #419 and #391. Both have the same root cause: on macOS nothing run under `srt` can enter raw mode.

## Root cause

Seatbelt matches `file-ioctl` rules by device path. The profile allows `file-ioctl` on `/dev/tty`, but a terminal is a pty slave (`/dev/ttysNNN`) that the alias does not cover. `TIOCSETA` returns `EPERM`, so the terminal stays in canonical mode and echoes the application's own protocol traffic as text: capability replies (#419), and KKP key encodings and mouse reports (#391).

## What changed

The profile now grants `file-ioctl` on the terminals the child inherits on stdin, stdout and stderr. This removes the need to enable `allowPty` for raw mode, which is the reason most people would turn it on.

| `allowPty` | grant |
|---|---|
| unset or `false` | `file-ioctl` on the inherited terminals |
| `true` | unchanged: every pty, for programs that allocate their own (tmux, script, node-pty) |

**Behaviour change:** `allowPty: false` used to emit no pty rule. It now behaves like unset.

The base profile already allows `file-ioctl` on `/dev/tty`, so the added surface is small. `TIOCSTI` stays denied in both modes, and a test checks that.

`wrapWithSandbox` returns a command string and the caller picks stdio afterwards, so terminals are resolved only when the caller passes the new `inheritsStdio` option. The CLI sets it. Library callers opt in.

## Tests

`test/sandbox/macos-pty-default.test.ts` covers profile generation, the resolver, and end-to-end behaviour through the real CLI. The end-to-end cases need a pty that is the child's controlling terminal, which Node cannot set up without a native dependency, so three small Python helpers do it. I also ran `vim` under the CLI by hand and it behaves normally.

<details>
<summary>How to try it</summary>

Run this in a terminal from the repo root:

```
echo '{"filesystem":{"denyRead":[],"allowWrite":["/tmp"],"denyWrite":[]},"network":{"allowedDomains":[],"deniedDomains":[]}}' > /tmp/s.json
bun src/cli.ts --settings /tmp/s.json -c 'stty raw; rc=$?; stty sane; echo rc=$rc'
```

On main it prints `stty: TIOCGETD: Operation not permitted` and `rc=1`. On this branch it prints `rc=0`.

</details>

## Docs

The README documents `allowPty` and fixes the violation-monitoring predicate, which matched the `sandbox-exec` process name and returned nothing.

macOS only. The Linux and Windows paths are untouched.


pcontrerasp:
@dylan-conway @ronleizrowice-ant can one of you take a look at this when you have time?

It fixes #419 and #391. On macOS nothing run under srt can enter raw mode, because the profile allows file-ioctl on /dev/tty but not on the pty slave the process holds. I rebased it onto main (v0.0.77) today. One behaviour change to know about: allowPty: false now behaves like unset.

I wrote it with Claude Code (Fable 5.1), reviewed it and tested it by hand on macOS 14 arm64. vim runs normally under the CLI and the e2e pty tests pass.

CI hasn't run on this PR yet. It needs a maintainer to trigger the workflow.

PR #504, opened 2026-08-29
linux: pass an over-long bwrap profile through --args

ronleizrowice-ant:
## Summary

`wrapCommandWithSandboxLinux` returns one string that the embedder runs as a single argument of `sh -c`, and Linux caps a single argument at 32 pages (128 KiB with 4 KiB pages). A sandbox profile with enough mounts, for example a tree with hundreds of nested repositories, goes past that. Then every sandboxed command in that tree fails at spawn with `E2BIG`, whatever the command is. This PR hands the mounts of such a profile to bubblewrap through `--args <fd>`. A profile that fits on the command line renders exactly as before.

## Example

With 137 KiB of mounts, `bwrap --ro-bind ... -- /bin/bash -c 'ls'` used to fail at spawn with `E2BIG`. Now the string is:

```
/bin/sh -c 'exec 9<"$1" && shift && exec "$@"' srt-args /proc/<wrapping pid>/fd/<n> \
  bwrap <options> --args 9 <options> -- /bin/bash -c 'ls'
```

It is still a simple command, so a prefix (`exec`, `timeout 30`) or a suffix (`&& next`) composes as before.

## Change

- **One place builds the string:** `renderBwrapInvocation()`. While the rendering stays 4 KiB under the cap it is `quote([bwrap, ...args])`, byte for byte what it was. The cap is read from `AT_PAGESZ` in `/proc/self/auxv`, so kernels with 64 KiB pages keep their 2 MiB. Past the threshold, only the filesystem mounts move to the file; the per-command environment and the command stay on the line.
- **The profile file has no name.** `openBwrapArgsProfile()` creates it with `O_TMPFILE` under `os.tmpdir()` (or `/dev/shm`), writes the mounts NUL-separated, and holds the descriptor open. `/bin/sh` opens it again through `/proc/<wrapping pid>/fd/<n>`, which gives a fresh read-only description at offset 0. Because the file is never linked into a directory, there is no path anyone could swap between the wrap and that open: not a command this process sandboxed, not another sandbox of the same user with the temp dir writable, not a rename above `TMPDIR`. Reaching it needs `/proc/<wrapping pid>`, and every sandbox this library starts has its own PID namespace and a fresh `/proc`.
- **Problems are refused at wrap time, with the reason,** instead of at spawn or inside bubblewrap: no directory accepts an `O_TMPFILE` file, or it cannot be reopened through `/proc`; more than bubblewrap's 9000 arguments; a NUL in a mount path; a command too long for one argument by itself. There is no named-file fallback. Such a profile failed with `E2BIG` before anyway.
- `cleanupBwrapMountPoints()` closes the descriptors, one per pending profile, refused renderings included. Nothing is named, so nothing is left on disk. The README documents all of this.

## Who sees a difference

- **A profile that fits:** nothing. No file, no descriptor, the same string.
- **A profile past the threshold:** it used to fail at spawn with `E2BIG`; now it runs.
- macOS and Windows are untouched. Not behind a flag.

## Known limits

- Another process of the same user, outside a sandbox, can read a pending profile through `/proc`. It can already read this process's memory, so that is nothing new.
- The string must run while the wrapping process is alive and before `cleanupAfterCommand()` releases the descriptor. After that the redirection fails and no command runs.
- The cap only exists because the argv is flattened into one string. A Linux `wrapWithSandboxArgv()` returning `[bwrap, ...args]` would avoid the file entirely. Not done here.

## Merge order

Fourth in the stack: #533, #514, #503, then this. #515 wants it: its extra binds halve a submodule-heavy repository's headroom.

## Testing

Linux x86-64, bubblewrap 0.11.2, unprivileged user namespaces, in `test/sandbox/linux-bwrap-args-file.test.ts`:

- the switch point, and a profile that fits naming no file;
- the file holds the mounts and only the mounts, and the string names nothing but `/proc/<pid>/fd/<n>`;
- one descriptor per pending profile, all released at cleanup (a refused rendering included); a string run after its cleanup runs no command;
- every refusal, the "no unnamed file possible" one in a fresh process with both candidate directories read-only;
- end to end: a profile past the cap applies its masks under `timeout 60 ... && echo AFTER`, and the command can reach neither fd 9 nor the wrapping process's `/proc`;
- end to end: the `TMPDIR` parent is renamed aside and a profile binding `/` read-write is planted at every path the string could still name; the command still runs under the profile it was wrapped with.

CI: seven of eight jobs green. The macOS arm64 job fails one proxy-attribution test (`commandId/commandText`) that this PR does not touch and that also fails intermittently on `main`.


sylvesterkaczmarek (commented):
The per-process `bwrapArgsDir` is mounted read-only into every sandbox, but read-only still exposes its contents. A long-running earlier sandbox can watch/list that directory and read a later invocation's pending profile before bwrap opens/unlinks it; those files include that later invocation's `--setenv` values and policy. The comment compares this to a sandbox seeing its own env/cmdline, but this is cross-invocation disclosure. Could the profile be handed through an already-open fd/memfd or otherwise not be shared readably between sandboxes?

sylvesterkaczmarek (changes requested):
Thanks for keeping the environment and command out of the shared profile file; that addresses the secret-value exposure I raised. I still see a policy-integrity blocker on `fe87913`.

The returned wrapper opens the args file by pathname at execution time, while `directoryIdentity()` checks it earlier, during wrapping. As the new README notes, a sandbox that can replace a writable ancestor of TMPDIR, or another runtime process's writable tmpdir, can substitute the pending profile before that open. A read-only bind in this process's sandbox does not protect those cases.

Please enforce a safe-storage precondition and refuse the file fallback when it cannot be met, or pass the profile through an immutable/open descriptor. A wrap-then-replace-then-execute regression should verify that replacement is rejected rather than consumed. This is a source review; I have not run a Linux reproduction.

sylvesterkaczmarek (approved):
Rechecked current `016a6687`. The pathname-substitution blocker is resolved: the over-long mount profile now lives in an unnamed `O_TMPFILE`, stays held by the wrapping process, and is reopened through `/proc/<pid>/fd/<n>` rather than by a replaceable pathname. The new tests cover the unnamed profile path. The lone macOS-arm64 CI failure is in an unrelated proxy-attribution test; Linux x86-64/arm64 and Docker jobs are green. No remaining blocker from my review.

dylan-conway (approved):

PR #506, opened 2026-08-29
Default write paths obey a read deny, and interior path spellings are canonicalised

ronleizrowice-ant:
## Summary

Three ways a filesystem rule stops meaning what it says. All three are fixed here.

1. **A convenience write path re-opens a read-denied directory.** Every write allowlist gets ten default paths. Eight are things the sandbox itself needs (the child's stdio, `/tmp/claude`); two are conveniences under the home directory that the caller never asked for, `~/.npm/_logs` and `~/.claude/debug`. Under `denyRead: ['~']` (or `['~/.npm']`, or a credential-file deny under `~/.claude`), Linux mounts a tmpfs over the denied directory and then binds the convenience directory back over it read-write. On macOS it stays writable under the read deny.
2. **A path with `//` or `/./` in the middle matches nothing.** `normalizePathForSandbox()` stripped only a *trailing* run of slashes, and did it before `~` was expanded. When the target does not exist yet, the usual case for a deny, an interior `//` or `/./` reached Seatbelt as written, and Seatbelt compares against the kernel's canonical path, so the rule denied nothing.
3. **An empty `$HOME` turns `allowWrite: ['~']` into the filesystem root.** `homedir()` is the empty string when `$HOME` is set and empty (node returns it verbatim), so `~` expanded to `''`, and `'' || '/'` made it `/`.

## Examples

| Config | Before | After |
|---|---|---|
| `denyRead: ['~/.npm']` | `cat ~/.npm/_logs/<host log>` prints the host's file, and a write there lands on the host | "No such file"; the write stays on the tmpfs |
| `denyRead: ['~/.aws//credentials']` (what `${dir}/${name}` gives when `dir` ends in `/`), file not created yet | denies nothing on macOS | denies `~/.aws/credentials` |
| `$HOME` set with a trailing slash | every `~` entry gets a `//` | canonical |
| `allowWrite: ['~']` with `HOME=` (set, empty) | `/` bound writable on Linux, `(subpath "/")` allowed on macOS | the empty spelling names nothing and is skipped |

## Change

**Default write paths.** `getDefaultWritePaths({ denyRead, allowRead })` leaves a home convenience directory out when a `denyRead` entry names it or a directory above it: the explicit deny wins over the implicit allow, and a caller who wants the directory writable lists it in `allowWrite`. An `allowRead` entry beneath the covering deny keeps it. Only those two paths are ever left out: the `/dev` nodes and `/tmp/claude` stay whatever is read-denied, so `denyRead: ['/']` with `allowRead` carve-outs behaves as before. Entries are compared as configured, not against the disk: `dir/**` counts as `dir`, and a glob covers a directory when it matches it or one above it, using the same `denyGlobRegex` the macOS profile is built from, now shared in `sandbox-utils.ts`.

**Path spellings.** After `~` and relative paths are expanded, and before the glob test, POSIX paths have runs of `/` collapsed (a leading `//` too) and `/./` components dropped, in one call for glob and non-glob spellings alike. A trailing `/.` and the trailing run are then dropped for non-glob spellings, where they are not semantic; a glob keeps its trailing slash and its `**` segments. The empty spelling is no longer rewritten into `/`, at the normaliser or at the Linux backend's own trailing-run strip, where it now falls out at the existence check. `/` itself is still `/`.

**Why not `path.normalize`.** It folds `..` on paper, which through a symlinked component aims the rule at a different file than the kernel would reach. An absolute spelling's `..` is left to realpath, and a path that still carries one when it reaches a backend gets a debug line saying the rule may not match, now for a glob spelling too, which previously said nothing at all.

## Who sees a difference

- **No read deny over either home directory:** nothing changes.
- **A read deny over one of them:** that directory is no longer writable. List it in `allowWrite` to keep it.
- **macOS deny rules with an interior `//` or `/./` over a path that does not exist yet:** they now match. Of the path-spelling work, that is the only change in what is enforced: bubblewrap tolerates the spelling and the Linux backend rebuilds its destinations, so that argv was already right.
- **An empty `$HOME`:** `~` entries no longer expand to the filesystem root on either backend.
- Every consumer that compares these paths by prefix now sees the canonical spelling, glob spellings included. Windows is untouched. Not behind a flag. The README gains a sentence on the two directories.

## Known limits

- On a case-insensitive volume, a deny spelled in a different case from the on-disk name does not cover the directory. Seatbelt matches case-sensitively, so that is consistent with what is enforced.
- **Folding `..` safely at the normaliser is not done here.** The fix is what `resolveSymlinkedDenyPath` already does on the Linux path: realpath the deepest existing ancestor and join the remainder. Until then such a rule matches nothing, with only a debug line to say so.
- **Nor the lexical fold of relative spellings.** `path.resolve` folds `../x` on paper before any of this runs, so a relative spelling through a symlinked component already aims elsewhere. Pre-existing, and it belongs with the item above.
- **Nor the Windows arm.** `C:\a\\b` over an absent target still reaches `expandWindowsFsPaths` as written, and a trailing separator there marks a directory for absent deny targets and must survive, so the collapse is POSIX-only.

## Merge order

`main` is merged in, #542, #514, #555, #536, #503, #553, #535 and #534 included, and nothing is waited on.

Both this branch and #542 touch `denyGlobRegex`, and the merge keeps two compilers rather than one. The string-taking one is exported from `sandbox-utils.ts` and is what a caller's configured spelling belongs in, every character of it being pattern; that is what `readRuleCovers` needs. The macOS file keeps an entry-taking `denyGlobEntryRegex` that splices #542's anchor back in as an escaped literal and calls the exported one for the tail, so the two cannot drift. Collapsing them onto the string helper would drop the anchor from every macOS deny regex and re-open the bypass #542 closed.

#536 and #503 land in the same file. They conflicted only in comment and test text around the rewritten `normalizePathForSandbox`; no code hunk conflicted. Where the changes meet, each checked on the merged tree:

- #503's read-deny expansion takes `getFsWriteConfig().allowOnly` as the set a deny must not be collapsed against. This branch narrows that set under a read deny, so a read-denied convenience directory is no longer counted as re-exposed there either, which only ever widens a deny.
- #503's glob walk takes a pattern's base directory from its static prefix, and now takes it from the collapsed spelling: `/data//x/*.pem` walks `/data/x` rather than `/data//x`, and every match it reports is spelled the way the comparisons downstream expect. The rule that refuses a pattern with no literal directory to start from is unchanged: `//*.pem` and `/./*.pem` are refused before the merge and after.
- #536 rejects a deny glob that keeps its trailing separator, deciding on the spelling as written. The collapse never touches a trailing separator, so each of its cases keeps its verdict: rooted slashed globs rejected (`/data/*/`, `/data/**/`, `/data/[ab]/`, `/data/?/`, a `denyWrite` and a `mode: "deny"` credential file), relative ones accepted (`path.resolve` strips the separator, so they are live rules), literals and allow globs untouched, and a slashed deny under `filesystem.disabled` accepted.
- #535 rebuilt what the Linux violation monitor is given: its write rules now come from `getFsWriteConfig()` rather than from `getDefaultWritePaths()` and `filesystem.allowWrite`. That getter is what this branch narrows, and the monitor is diagnostic only, so the merged call keeps #535's derivation and its built-in denies and adds the unnarrowed default write paths back to the allow list whenever the filesystem policy is on: a write into a convenience directory this branch drops lands on the read deny's writable tmpfs, so bubblewrap permits it and it is not a violation. With `filesystem.disabled` the handed-over lists are exactly #535's, `/` and no built-in denies.
- `sandbox-manager.ts` imported `normalizePathForSandbox` twice from `sandbox-utils.ts`, which `tsc` rejects with TS2300. The duplicate is dropped in its own commit.

## Testing

Re-measured against `main` at the merge, so under #503's collapsed read-deny globs and its mounts at the location a deny path resolves to.

Closest to real: end to end on Linux under real bubblewrap, with host content in both convenience directories. Under `denyRead: ['~/.npm']`, `main` prints the host log and its write lands on the host; this branch prints "No such file", the write fails inside the sandbox (the tmpfs over `~/.npm` carries no `_logs` directory) and nothing reaches the host, and the directory is no longer bound in the argv. `~/.claude/debug`, which that deny does not cover, stays readable and writable. The macOS side was checked under real `sandbox-exec` with the emitted profile, from a working directory named `a[b/c]d`: its own `.git/hooks`, a nested repository's hooks and the caller's relative `denyWrite` are each refused with `Operation not permitted`, an ordinary file in the project is written, and `<root>/abd`, the sibling the bracket class would have matched instead, is untouched by the deny. A second `sandbox-exec` pair takes a `denyWrite` spelled `<root>/area//secret.txt` over a file that does not exist yet: `main` writes it, this branch refuses it.

Two new suites cover the rest, both driving `getFsWriteConfig()` and a wrapped command in a child process with its own `$HOME`, since `$HOME` is read once at start-up. A differential over 366 spellings, under four `$HOME` values and with and without the literal option, shows the merged normaliser identical to this branch's before the merge for every one of them; against `main` it differs on exactly the spellings this fixes, an interior `//` or `/./`, a leading `//`, a trailing `/.`, and every `~` spelling when `$HOME` itself ends in a slash. Full suite on Linux: 1501 pass, 17 fail; the same 17 as a clean run of `main` on the same host (the known `--control-fd`, seccomp-binary and proxy flakes), none attributable to this change.

- `test/sandbox/default-write-paths.test.ts` (26 cases on Linux): every way of naming the covering deny (`<home>`, `~/`, `~/**`, `/`, `/**`, a glob over the parent, one whose match is the directory itself), siblings that must not match, the `allowRead` cases, and an invariant that nothing under a read-denied home survives, so a third home default added later cannot bring the leak back.
- `test/sandbox/macos-literal-deny-brackets.test.ts` gains three cases for the merge itself: under a bracketed working directory every regex the profile emits escapes the brackets, and the mandatory subtree patterns are anchored at that directory, so a nested repository's hooks are covered and the sibling the bracket class would have matched is not; and a caller's own glob deny compiles to exactly what the exported string compiler returns, so the entry-taking wrapper cannot drift from it. The first two fail with the anchor dropped, alongside two of #542's.
- `test/sandbox/path-canonicalization.test.ts`: the collapsed spellings, `..` kept as written (glob and non-glob), the dot-file segments the `/.` drop must not touch, the debug line a glob with an unfoldable `..` now emits, the macOS profile for an absent `//` deny and for a glob spelled with `//`, and both `$HOME` cases, a trailing slash and an empty one. The absent base is a guaranteed-absent child of a `mkdtemp` directory rather than a root path assumed never to exist.
- The empty-`$HOME` arm asserts the invariant on whatever runtime runs it, and only bites where `homedir()` honours an empty `$HOME`: node returns `''`, bun falls back to the passwd entry. Measured on the compiled build under node: before, `~` normalises to `/` and the root is bound writable; after, it stays empty and is skipped.

</details>

<details><summary>Further detail</summary>

- `denyRead: ['~/proj/../.aws/*.pem']` with `~/.aws` absent compiles to a regex holding `/../` and matches nothing, silently. It still compiles to the same regex; what is new is a debug line naming the spelling that reached the backend. The line no longer claims the path "does not exist", because the same branch is taken for one that could not be resolved for any other reason, and a glob spelling now emits it at all, having returned before the realpath attempt.
- `getFsWriteConfig()` stays a plain getter. It and `wrapWithSandbox()` share one helper, which also counts the credential-file denies. Called with no arguments, `getDefaultWritePaths()` is still the constant list.
- The comment above the trailing strip is corrected: a Seatbelt `subpath` filter is the one place where `dir` and `dir/` are interchangeable.

</details>


sylvesterkaczmarek (commented):
Could `getDefaultWritePaths()` account for denyRead globs too? On Linux, `~/.npm/*` is expanded later to a concrete `_logs` deny, but this helper ignores non-trailing globs and still adds `_logs` to `allowOnly`, which can bind it back over the read deny. That seems to leave the same fail-open path for glob-based policies.

sylvesterkaczmarek (commented):
Focused re-review of `365b95f`: configured glob denies now reach `readRuleCovers()`, including the `~/.npm/*` case I raised. The write-config getter and command wrapper use the same pure read-policy helper, without triggering credential masking or glob enumeration in the getter. This resolves my original concern. The focused suite passes on macOS: 21 passed, 4 Linux-only cases skipped. Thanks for the follow-up.

sylvesterkaczmarek (approved):
Re-reviewed current 0ef30e21. Interior slash and /./ canonicalization now applies after expansion to both ordinary paths and glob spellings on POSIX, while preserving trailing-glob semantics and leaving .. to realpath. The added regressions cover clean equivalence and the HOME trailing-slash case. CI is green. No remaining blocker from me.

sylvesterkaczmarek (approved):
Re-reviewed current d2fdaabe. The follow-up keeps the POSIX canonicalization centralized, covers glob spellings, reports unresolved parent references, and fixes the empty-HOME path so it cannot become a root write grant. The platform CI matrix is green. No remaining blocker from me.

dylan-conway (approved):

PR #515, opened 2026-09-02
Mandatory denies: submodule git dirs, .git pointer files, nested hooks, gitignored paths

ronleizrowice-ant:
## Summary

The mandatory write denies exist so that a sandboxed command cannot leave behind something the host's own tools will later execute: a git hook, or a `core.fsmonitor` line in a git config. For git they covered the working directory's `.git/hooks` and `.git/config`, plus nested repositories found by a scan (Linux) or a pattern (macOS). Seven shapes slipped through with the default configuration, each confirmed under bubblewrap or Seatbelt. This PR closes them, reads a `.git` pointer file the way git reads it, and makes discovery fail closed - while keeping the other half of that bargain: a sandboxed command must not be able to plant a refusal that every later command in the tree hits and nothing inside the sandbox can lift. On macOS every path it finds on disk is denied as a literal entry and every pattern it writes is anchored at the working directory, so a submodule or a working directory whose own name contains `[` or `]` is covered like any other (#542).

## What slipped through

| Shape | Platform | Problem | Now |
|---|---|---|---|
| Nested repository at scan depth | Linux | The default depth of 3 reaches `pkg/.git/config` but not the hook files one level deeper, so a nested repository's hooks stayed writable. | Any regular file directly inside a `.git/` directory marks a repository. Not keyed on `HEAD`, which a command could move aside. See **Finding a nested repository**. |
| Ignore files and the ripgrep config | Linux | ripgrep honours `.gitignore`/`.ignore`, which the sandboxed command can write, so one command could hide `pkg/.git` from the next command's scan. `RIPGREP_CONFIG_PATH` routinely points inside a project too, and one `--max-filesize=1` line in it empties the scan with no error. | The scan runs with `--no-ignore --no-config` and NUL-delimited output. |
| A scan that fails | Linux | An error from the scan was logged and swallowed, so a command that left one directory unreadable had the next one sandboxed with no nested denies at all. | See **Failing closed**: what the run can be carried on from is decided by the `(os error N)` codes it printed, and never by the paths printed beside them. |
| Submodules | both | A submodule's git directory is `.git/modules/<name>/`. Its `hooks/` and `config` matched nothing. | Linux walks `.git/modules` for the cwd's repository and every nested one. macOS lists the cwd's own as literal paths and matches nested ones by pattern. |
| `.git` pointer files | both | A worktree or submodule checkout has a `.git` *file* containing `gitdir: <path>`. Rewriting it hands the host's git a directory the command prepared. | The file is read-only and cannot be removed or renamed over, and the git directories it leads to are denied too. Creating one where none exists (`git worktree add`) still works. |
| `commondir`, `config.worktree` | both | `commondir` redirects where git reads `hooks/` and `config` from. `config.worktree` is read instead of `config` when `extensions.worktreeConfig` is on. | Both are denied in every git directory (`config.worktree` follows `allowGitConfig`; `commondir` always). Denying a file that is not there means mounting something at it, and git refuses to run at all against a `commondir` it cannot read - which both `/dev/null` and an empty file are, on the host as much as in the sandbox. So on Linux the wrap writes the mount point itself, holding what git concludes with no file there, and binds a read-only copy of the same bytes over it. |
| Two-segment names | Linux | A nested `.claude/commands` within scan depth got one bind per existing file, so new files could still be created. | The directory itself is denied. |

## How a pointer and a `commondir` are read

- **Byte for byte, as git's `read_gitfile_gently()` does.** A differential test against real git is what holds this (see Testing).
- **A target that is not a git directory is not followed. One that does not exist yet is denied anyway,** so the command cannot create it and fill it with hooks first.
- **`..` after a symlink:** `link/../x` does not land where folding the string on paper says, so both the lexical answer and the kernel's are computed, and where they differ both directories are denied.
- **A target this process cannot read is denied whole,** at its deepest reachable ancestor. git, running as the same user, cannot read through it either, so that repository is already broken for everyone; refusing instead would be a brick one `chmod` plants.
- **Where git's answer cannot be known at all, the wrap is refused** - see **Refusals a command can plant**.

## Failing closed

ripgrep lists the rest of the tree and exits non-zero when an entry gets in its way, so a failed run still carries matches. Whether they can be used is decided by the `(os error N)` codes on its stderr, never by the paths printed beside them: a directory's name is chosen by whoever created it and reaches stderr unescaped, so a name holding `: `, a trailing space or a newline would otherwise pick the path that gets denied. A name can only add codes, and an unrecognised code refuses.

- every code `ENOENT` - entries that went away while the scan walked. There is nothing at those paths to deny, and mounting one would plant a file on the host exactly where the command deleted one. The listing still counts.
- any code `EACCES` - a directory it could not read holds an unknown tree, so it is denied whole and the listing still counts. WHICH directories is settled by a bounded walk of the working directory, to the scan's own depth, skipping `node_modules` as the scan does, and inside what is left of the scan's own time budget.
- anything else, no code at all, or an `EACCES` the walk finds no directory for: the command is refused with `deny_scan_failed`, and the raw stderr (truncated) is in the message.

A scan that could not be run at all - the binary could not be spawned, which is the dependency the start-up check refuses on, met later - and one that does not finish inside its timeout are refused the same way: the command about to run is the one that could have made the tree slow to walk. An abort is the caller's own answer and comes back as the `AbortError` it is. A run killed by something that is not the timeout is reported as the signal it was, not as a timeout, and still refuses.

The same whole-directory deny stands in elsewhere: for a directory the `.git/modules` walk could not list, and, once per branch, for what the walk would have descended into at `MAX_SUBMODULE_WALK_DEPTH` - the `modules` directory beneath the git directory it stopped at, or that directory itself when it is not a git directory. Denying that git directory outright would take its `objects`, `refs` and `index` with it and stop git working in the submodule.

A git directory that takes no mount point at all - one this process cannot write, such as a root-owned vendored checkout or a read-only mount - is bound read-only whole rather than refusing. It needs nothing written to the host, costs what was already unwritable, and is what stops the command from making the directory writable and creating the redirect file itself. `deny_placeholder_unavailable` is left for the cases where the host genuinely cannot take the file: no space, or no writable temporary directory for the placeholder.

The one consumer that must not be refused is the violation monitor, started once for a session to judge a refused write against the same cwd denies. It asks through a wrapper that falls back to this directory's plain deny paths and warns, so one unreadable repository does not fail the whole start-up; every wrap in it is still refused.

## Refusals a command can plant

Two wrap-time refusals follow from a file a sandboxed command is allowed to write, and the command that would put the file right is refused as well, because every wrap in that working directory hits the same file. **Both are lifted only by a command run outside the sandbox**, and both name the offending path, or the line the scan printed, in the message so that the operator knows which:

- `deny_git_metadata_unreadable` - a `.git` pointer or a `commondir` whose target cannot be worked out the way git works it out: bytes that are not valid UTF-8, a `commondir` larger than the size read here, or one that is there and cannot be read. Creating a new `.git` pointer inside an allowed write path is deliberately allowed, so `printf 'gitdir: \xff' > sub/.git` is one command. On macOS the same condition is a `GitMetadataError`, which now carries a name and a `.code` and is exported from the package root.
- `deny_scan_failed` for a scan failure no deny stands in for, as above.

Everything else a command can do to the tree is answered with a deny rather than a refusal: an unreadable directory is denied whole, an unwritable git directory is denied whole, and a vanished file is nothing.

## Finding a nested repository

A nested repository is recognised by any regular file directly inside its `.git`, which is why it is found at the depth the repository sits at rather than one level further down where the hook files are. With `allowGitConfig` off - the default - the deny of `config` keeps such a file there and it cannot be removed from inside the sandbox, so one command cannot empty the directory and hide the repository from the next command's scan. With `allowGitConfig` on it could, so each command additionally walks the working directory to the same depth (the same walk the `EACCES` case uses) and denies the `.git` directories it finds. That walk is the cost of allowing config writes and is not run otherwise.

## Who sees a difference

- **An ordinary repository** keeps its index, objects, refs and `git commit`, which a test holds by committing inside the sandbox twice over the same repository. It does see one change on Linux, and on the host as much as in the sandbox: for as long as a wrapped command runs, its git directory holds a `commondir` file containing `.` (and so does each submodule's), written by the wrap before bubblewrap starts and removed after the command. `.` is what git concludes with no `commondir` there, so every git command keeps working - but the file's existence is not nothing: `git rev-parse --git-common-dir` and `--git-path` print the absolute real path where they printed a relative one. A process killed before its cleanup leaves the file behind; git reads it happily, and the next wrap in that repository takes it away. An empty `commondir`, which git refuses outright and which is what older releases left, is repaired the same way - unless it still has write bits, which means a host `git` is between creating the file and writing it, and then it is left exactly as found and only covered for the command. The exception is a tree that trips a whole-directory deny: it is read-only inside the sandbox, `objects`, `refs` and `index` included. Scan depth and Windows are unchanged. Not behind a flag.
- **A command whose scan fails** is refused rather than wrapped with the working directory's own denies alone: the scan is where every nested repository below it comes from, so a command that blinds it - one `chmod` of one directory was enough - would otherwise leave the next command a tree of writable hooks. The failures that still wrap are the two above, and neither takes a path from what the run printed.
- **A caller reading `.code`** gets one new `LinuxSandboxProfileError` code, `deny_git_metadata_unreadable`, and `GitMetadataError` from the package root on macOS. `RipgrepConfig.timeoutMs` is now in `RipgrepConfigSchema`, where the type already had it.
- **Git operations these denies break** (listed in the README): `rm -rf` or `git clean -ffdx` over a submodule checkout or linked worktree; `git worktree remove`/`move`/`repair`; `git submodule deinit`; `git submodule update --init` for a submodule not yet cloned; `git push -u` and `git checkout -b x origin/y` from a linked-worktree cwd; `git init`/`git clone` into a subdirectory.
- **Linux command-line cost:** a git directory now costs up to four read-only binds instead of two, each submodule is one git directory, #514 pins each of those directories on top of that, and a wrap that mounts a redirect placeholder adds one more bind for the store it comes from. The 128 KiB `execve` limit that put the ceiling near 240 submodules, past which every sandboxed command failed with `E2BIG`, no longer applies: #504 has moved the mounts off the command line, leaving bubblewrap's own limit of 9000 arguments. Re-measured on this head, with the walk and the whole-directory denies in it: still 594 submodules, and at 595 the wrap is refused up front with `too_many_arguments` and a message naming the limit, rather than failing at exec time. Building that 594-submodule profile takes about 210 ms, the same as before writing the redirect mount points here rather than leaving them to bubblewrap.

## Known limits

A pointer file or a pattern-matched path is protected where it is, so a command could rename the directory *holding* it and put a fresh one in its place (`mv lib lib.old && mkdir lib && echo 'gitdir: ...' > lib/.git`). On macOS that is still open for the pattern denies; the literal ones are safe from it. On Linux it is closed for everything the scan reached, now that #514 is in: the ancestors of every denied path are pinned, and on the merged tree under bubblewrap `mv` of the directory holding a denied pointer file, and of the package directory above a nested repository's hooks, both fail with `EBUSY`.

The scan never looks inside a `node_modules`, so a repository under one is never found and its hooks stay writable. That exclusion is permanent: the scan walks gitignored data as it is, and a `node_modules` is where the cost of that would land.

A tool that reads the mere existence of a `commondir` as "this is a linked worktree" misreads an ordinary repository while a wrapped command runs there, and until the next wrap if a killed process left the file. No content avoids that - the alternative is a file git cannot read.

A `commondir` holding exactly `.` is recognised as this wrapper's own by its content alone, so one a user put there by hand is removed by the next wrap's cleanup along with the wrapper's.

Cleanup deferral is per process. Two processes wrapping commands in one repository can each claim the same mount point, and the one that cleans up first removes it while the other's sandbox is still running, which detaches that bind and stops that one deny applying inside it. Binding the placeholder from the store rather than from the mount point keeps the sandbox starting; the deny is what is lost.

The `.git/modules` walk follows a symlinked entry, because git does. That has three consequences, none of them an escalation (each is a command making something read-only for the commands that follow it, never writable) and all three are follow-ups.

<details><summary>The three symlink consequences</summary>

- A symlinked entry at the walk's depth bound is denied under the path as written, which on Linux the backend resolves to the link's target. A command that can write under `.git/modules` can therefore point one directory of its choosing at itself and have it mounted read-only for the next command. Before this PR the reachable set was four file paths, and only where the target already held a git marker. Hardening the walk against symlinks is the fix.
- A symlink loop under `.git/modules` cannot be stat'ed (ELOOP), and the deepest ancestor the walk can reach for it is `.git/modules` itself, so the whole tree is denied whole.
- The walk's breadth is unbounded: only its depth is capped. A very wide `.git/modules` costs one `readdir` per directory.

</details>

## Merge order

`main` is merged in at a7f016b, so this sits on top of every change to the Linux deny loop and its producer that had landed by then: #542, #514, #503, #555, #536, #553, #535, #534, #562, #524 and #506. Nothing is owed ahead of it, and GitHub reports no conflict with `main` as it stands.

## Testing

Closest to real: `test/sandbox/mandatory-deny-paths.test.ts` drives each shape through the real wrapper, under bubblewrap 0.11 on Linux and Seatbelt on macOS, with a positive control wherever a failure alone would prove nothing. The parity claim is held by `test/sandbox/git-pointer-parity.test.ts`, a differential test against real git: each pointer is resolved twice, through this code and through `git rev-parse --absolute-git-dir` / `--git-common-dir`. Wherever git follows a pointer, that directory's `hooks`, `config`, `config.worktree` and `commondir` must be denied, or the wrap must have been refused; denying more passes, denying less fails. Also driven by hand through the `srt` CLI on Linux and `sandbox-exec` on macOS.

- Parse edges, unit-tested: both sides of the 1 MiB bound, NUL truncation, CRLF, every refusal path, `..` after a symlink, loops, FIFOs, the walk's depth bound.
- The parity corpus crosses prefix, trailer and path spellings with target shapes, plus 150 seeded random mixes.
- The depth bound is covered by one parameterised case asserting the whole `{ gitDirs, unreadableDirs }` shape, with a plain directory and a git directory at the bound, its segment count derived from `MAX_SUBMODULE_WALK_DEPTH`; plus a Linux arm that builds the argv and, where bubblewrap can create its namespaces, runs the wrapped command: inside a git directory at the bound a write to `objects/` succeeds while `hooks/`, `config` and creating `modules/` fail.
- The scan's failure paths are driven by a fake `rg` that writes rg's own stderr byte for byte, so they do not turn on the uid: a run whose codes are all `ENOENT` carries on and mounts nothing at the vanished path; one reporting `EACCES` with nothing unreadable on disk is refused; one with an error code no deny stands in for, and one that says nothing at all, are refused with the raw line in the message; one outlives a short `timeoutMs`; one cannot be spawned; and an aborted signal comes back as the `AbortError` it is. Where the uid allows a directory to be unreadable, the same fake drives the name shapes that defeat reading paths out of stderr - a directory named `src: x` holding an unreadable `locked`, one named `.: x`, a name ending in a space, a name holding a newline, and a stderr mixing `ENOENT` with `EACCES` - and each time the unreadable directory is denied and nothing beside it is. The real `rg` closes the loop under bubblewrap: one wrapped command takes a directory named `a: b` to mode 000, and the next is still refused the nested repository beside it and refused the `chmod` that would open it again.
- The files git reads back have argument-level cases for a `commondir` absent, present, left empty by bubblewrap's own shape, left empty with write bits by a host git mid-write, and already holding the placeholder; for a caller's own deny of the same file spelled relatively and with a trailing slash; for an empty `config.worktree` (legitimate, so left alone unless it has bubblewrap's shape); for what the cleanup removes and what it must not; for a git directory at mode 0555, which is denied whole instead of refusing; for a wrap that throws after the files were written, which leaves nothing behind; and for the refusal when no placeholder can be written at all. The killed-wrap case is driven by a child process that wraps and then `SIGKILL`s itself, so the leftover is the wrapper's own rather than a hand-made file.
- Live arms under bubblewrap, on a repository with a submodule and a linked worktree: while a wrapped command runs, the host's `git status`, `log`, `worktree list`, `rev-parse --git-common-dir`, a commit and `git status` in the submodule all succeed, and both `commondir` files hold `.`; after the command the files are gone and git still works. A leftover `.` is read by git and removed by the next wrap; the empty `commondir` an older release leaves makes git refuse every command and is repaired by the next wrap. With `$TMPDIR` inside an allowed write path, `chmod` and an in-place write of the placeholder source both fail inside the sandbox with "Read-only file system" and git still resolves its own git directory; twenty wraps mint one store directory.
- `test/utils/ripgrep.test.ts` covers the NUL split, what a failed run carries back, that `--no-config` is on every run, and that a kill inside the time the run was given is reported as the signal it was rather than as a timeout. A unit case covers the violation monitor's fallback: a repository whose `commondir` is past the size read refuses every wrap, but leaves the monitor the cwd's plain deny paths - and no deny under a `.git` that is not there.
- The macOS bracket cases sit with #542's own in `test/sandbox/macos-literal-deny-brackets.test.ts`, which is why that file appears in this diff: under a working directory named `a[b/c]d`, a submodule git directory, a directory the modules walk stopped at, and a pointer's target and its `commondir`'s target are each denied by subpath; no emitted regex carries a bracket that came off the filesystem; and the patterns with no path to enumerate, a nested repository's submodule and the `.git` pointer filter, are anchored at the escaped working directory. Three macOS-only cases run the profile under real `sandbox-exec`: a hook planted in a bracketed submodule git directory is refused, so is a write in the directory the walk stopped at, and a write elsewhere in that git directory still succeeds.

</details>

<details><summary>How the pointer parse matches git byte for byte</summary>

A regular file only (opened `O_NONBLOCK` and type-checked on the fd, so a FIFO cannot block the host), at most the 1 MiB git accepts, the 8-byte prefix at offset 0, trailing `\n`/`\r` stripped, the path ending at the first NUL. `commondir` is parsed the same way minus the prefix. For the `..`-after-a-symlink case, the lexical answer is `gitMetadataTargets` and the kernel's is `physicalPath`.

</details>


sylvesterkaczmarek (commented):
One linked-worktree gap remains: `gitFileDenyPaths()` only denies `config.worktree` when that file already exists. If the common config already has `extensions.worktreeConfig=true`, Git will consult a subsequently created `.git/worktrees/<id>/config.worktree`; a writable repo can then create it after wrapping and set executable config such as `core.hooksPath`. Could creation of `config.worktree` be blocked whenever worktree config is enabled, rather than only protecting a pre-existing file?

sylvesterkaczmarek (changes requested):
Thanks for addressing the `config.worktree` creation gap. One new bypass remains on `af8a067` in `readGitMetadataFile()`: reaching its 8 KiB read limit causes `gitFileDenyPaths()` to protect only the pointer, although Git still accepts that file.

Reproduced on macOS with real Git and Seatbelt: create a valid `gitdir: <target>` pointer, then append 9,000 newline bytes. `git rev-parse --absolute-git-dir` succeeds both before and after padding. With the ordinary pointer, a sandboxed harmless write to the target gitdir/config gets EPERM; with the padded pointer, the same write succeeds. The target hooks/config denies disappeared.

Please read metadata within Git's accepted bound, or fail closed when the pointer cannot be fully validated, and add this padded-pointer regression. The existing mandatory-deny suite passes locally (65 passed, 21 Linux-only skips), so it does not catch this case.

sylvesterkaczmarek (approved):
Re-reviewed the current head. The padded-pointer bypass from my previous changes-requested review is addressed: .git metadata is read through Git's accepted 1 MiB bound, oversized commondir fails closed, and the differential parity suite now exercises the 9,000-newline case, 64 KiB padding, the exact 1 MiB boundary, and 1 MiB + 1 against real Git. The platform CI matrix is green. No further blocking findings from me.

sylvesterkaczmarek (approved):
Re-reviewed current d258cc57. The depth-bound handling now denies the directory where the .git/modules walk stops, and the regression covers a deeper git directory. This is consistent with the existing unreadable-directory behavior. CI is green. No remaining blocker from me.

sylvesterkaczmarek (approved):
Re-reviewed current 8fad968c. At the depth bound, a recognized git directory now keeps its ordinary git state writable while its nested modules subtree is denied whole; its own hooks and config remain protected separately. The added Linux and macOS regressions cover that boundary, and CI is green. No remaining blocker from me.

sylvesterkaczmarek (approved):
Re-reviewed current `d843decf`. The latest PR commit is regression coverage only: it exercises bracketed submodule git directories, pointer/commondir targets, and the literal subpath denies on macOS without changing the mandatory-deny implementation I previously approved. The cross-platform test matrix is green. No blocker from me.

dylan-conway (dismissed):

sylvesterkaczmarek (approved):
Re-reviewed current `e275b42b` after the recent Linux redirect-placeholder changes and main merges. The new host-visible Git placeholder path remains fail-closed: absent redirect files are prepared without overwriting an existing redirect, the bound placeholder source lives in the isolated store and is pinned read-only after writable mounts, cleanup only removes wrapper-owned destinations while their bytes still match, and an unusable placeholder refuses the wrap. The monitor fallback changes only violation attribution; command wrapping still rejects unreadable Git metadata. The cross-platform CI matrix is green. No new blocker from me.

PR #566, opened 2026-09-18
ci: run the workflows on Bun 1.4.2

dylan-conway:
## Summary

Pin Bun 1.4.2 in both workflows. The test workflow pinned 1.3.14; the release workflow pinned 1.3.14 in two jobs and 1.3.1 in one.

Bun 1.4 replaced large parts of the Node compatibility layer the tests exercise — the `node:http` client and `child_process` stdio among them — so a pass or a failure on 1.3.x says little about the runtime the package is used with.

In the release workflow Bun only runs the vendor build scripts (`build:seccomp`, `build:java-agent`, `build:srt-win`). The test workflow runs the same three scripts, so this pull request's run covers them on the new version.

## Test fix

On 1.4.2 one test failed on every Linux and macOS leg: `proxy-deny-violations.test.ts › uses the default policy reason when filterRequest denies without one` timed out at 5s. It waits for the client socket's `'close'` without ever reading the socket. A socket nobody reads never reaches `'end'`, so it never emits `'close'` — that is Node's behaviour (the same pattern hangs under Node 22) and Bun's from 1.4; Bun 1.3 emitted `'close'` regardless, which is the only reason the test passed. Its two sibling tests already attach a `'data'` listener; this one now calls `sock.resume()`.

## Tests

The run on this pull request is the test: the full matrix on Bun 1.4.2.


dylan-conway:
Superseded by #565, which carries the same Bun 1.4.2 pins and the proxy-deny-violations test fix.

PR #569, opened 2026-09-18
Mandatory denies: submodule git dirs, .git pointer files, nested hooks, gitignored paths

ronleizrowice-ant:
## Summary

The mandatory write denies exist so that a sandboxed command cannot leave behind something the host's own tools will later execute: a git hook, or a `core.fsmonitor` line in a git config. For git they covered the working directory's `.git/hooks` and `.git/config`, plus nested repositories found by a scan (Linux) or a pattern (macOS). Eight shapes slipped through with the default configuration, each confirmed under bubblewrap or Seatbelt. This PR closes them, reads a `.git` pointer file the way git reads it, and makes discovery fail closed - while keeping the other half of that bargain: a sandboxed command must not be able to plant a refusal that every later command in the tree hits and nothing inside the sandbox can lift. On macOS every path it finds on disk is denied as a literal entry and every pattern it writes is anchored at the working directory, so a submodule or a working directory whose own name contains `[` or `]` is covered like any other (#542).

## Latest round (after 4750e50)

Eight changes, each with a test that fails at 4750e50. Four are the branch's own new cover being switched off, or turned into a lockout, by a few ordinary writes of a wrapped command:

- **A deny that lands on the work tree, however it is spelled.** The guard that keeps a whole-directory deny off the working directory and the write roots judged each deny path as written. A git directory entry that is itself a link to the checkout (`nested/.git/hooks -> ../..`) is spelled inside the git directory, passed the guard, and was bound where it lands: the whole project read-only from the next command on, with nothing inside the sandbox able to undo it. Every deny path is now judged by where it lands. What a degrade adds past the argument cap was never guarded at all and now is; a bind taken back there leaves the precise denies it stood for.
- **`deny_unresolvable_git_entry` only for the working directory's own repository.** A looping chain behind a nested repository a command made, or behind a `.git/modules` that was not there, refused every later command. It now gets the warning a looping pointer value already got, for the same reason. See **Refusals a command can plant**.
- **The `.git/modules` walk does not stop at a directory that looks like a git directory.** `touch .git/modules/vendor/HEAD` made `vendor` a git directory and `vendor/lib` was never listed again, on both platforms. The directory is recorded and its other children are walked too, except what git itself keeps there.
- **A git directory by where it is.** `mv .git/worktrees/wt/HEAD aside` made the next wrap deny only the pointer file and skip the `commondir` behind it. `.git/worktrees/<id>` and anything under a `.git/modules` is a git directory whatever it holds, judged on the path the kernel lands on; elsewhere `commondir` counts as a marker beside `HEAD` and `objects`.
- **`.git/worktrees/*` is listed from the repository that keeps it.** After `git worktree add ../wt` the pointer file is outside the write roots, but `<cwd>/.git/worktrees/wt/commondir` was an ordinary writable file. `commondir`, `gitdir` and `config.worktree` of every linked worktree are denied, on both backends, and they degrade like submodules. The same holds for the linked worktrees a SUBMODULE's git directory keeps (`.git/modules/<name>/worktrees/<id>/`, where `git worktree add` inside a submodule puts one): they join that submodule's deny paths, so the one bind a degrade puts on its git directory covers them.
- **A nested git directory with mode 0555** made every wrap in the tree die at bubblewrap start-up: the directory went whole, and a placeholder for another absent entry of it was still emitted after the bind. Buffered denies under a git directory that goes whole are now dropped, in both orders.
- **The placeholder repair writes through the file it checked.** It checked the name with `lstat` and then wrote and changed mode by path, in a directory a running command may be writing. It now opens once without following links, checks the descriptor, and writes through it; the read-back and the cleanup follow no link either.
- **Two unbounded loops are bounded.** The profile rebuild stepped one submodule per rebuild where a step gave no words back (402 builds and 20 s for 400 submodules outside the write roots; now 3). Symlink-chain and pointer-value resolution ran outside every clock check; it now spends the same budget as the walk and ends in the same `deny_scan_failed`. The walk that goes on past a git directory left a link at `config`, `commondir` or `config.worktree` to the git directory that owns it rather than following it a second time as a possible submodule: that was seven walks of a chain per git directory where four were needed, and it took a synthetic long-chain layout that used to wrap in about seven seconds past the ten-second budget.

Not done, and listed under **Known limits**: requiring more than one marker for a git directory outside a git-owned location, listing sibling worktrees from a working directory that is itself a linked worktree, and walking the `modules` of a git directory a pointer names.

## What slipped through

| Shape | Platform | Problem | Now |
|---|---|---|---|
| Nested repository at scan depth | Linux | The default depth of 3 reaches `pkg/.git/config` but not the hook files one level deeper, so a nested repository's hooks stayed writable. | Any regular file directly inside a `.git/` directory marks a repository. Not keyed on `HEAD`, which a command could move aside. See **Finding a nested repository**. |
| Ignore files and the ripgrep config | Linux | ripgrep honours `.gitignore`/`.ignore`, which the sandboxed command can write, so one command could hide `pkg/.git` from the next command's scan. `RIPGREP_CONFIG_PATH` routinely points inside a project too, and one `--max-filesize=1` line in it empties the scan with no error. | The scan runs with `--no-ignore --no-config` and NUL-delimited output. |
| A scan that fails | Linux | An error from the scan was logged and swallowed, so a command that left one directory unreadable had the next one sandboxed with no nested denies at all. | See **Failing closed**: what the run can be carried on from is decided by the `(os error N)` codes it printed, and never by the paths printed beside them. |
| Submodules | both | A submodule's git directory is `.git/modules/<name>/`. Its `hooks/` and `config` matched nothing. | Linux walks `.git/modules` for the cwd's repository and every nested one. macOS lists the cwd's own as literal paths and matches nested ones by pattern. The walk has no depth bound - a submodule's name is its path and submodules nest, so how deep a tree goes is a question about the repository - and what bounds it is time, as much as the scan is given. Past what bubblewrap takes arguments for, the denies degrade rather than refuse the repository: see **Submodules past the argument cap**. |
| `.git` pointer files | both | A worktree or submodule checkout has a `.git` *file* containing `gitdir: <path>`. Rewriting it hands the host's git a directory the command prepared. | The file is read-only and cannot be removed or renamed over, and the git directories it leads to are denied too. Creating one where none exists (`git worktree add`) still works. |
| `commondir`, `config.worktree` | both | `commondir` redirects where git reads `hooks/` and `config` from. `config.worktree` is read instead of `config` when `extensions.worktreeConfig` is on. | Both are denied in every git directory (`config.worktree` follows `allowGitConfig`; `commondir` always), a linked worktree's under `.git/worktrees/<id>/` included, which is reached from the repository that keeps it as well as from the worktree's own `.git` file and has `gitdir` denied beside the two. Denying a file that is not there means mounting something at it, and git refuses to run at all against a `commondir` it cannot read - which both `/dev/null` and an empty file are, on the host as much as in the sandbox. So on Linux the wrap writes the mount point itself, holding what git concludes with no file there, and binds a read-only copy of the same bytes over it. |
| Two-segment names | Linux | A nested `.claude/commands` within scan depth got one bind per existing file, so new files could still be created. | The directory itself is denied. |
| A git-directory entry that is a symlink | both | `hooks`, `config`, `commondir`, `config.worktree` or an entry of a `.git/modules` can be a link. A deny covers one half of one: on Linux the bind lands on the target and the link is left replaceable in a writable directory; on macOS the filter covers the link and not what it leads to. Either way the host's git between two commands runs what was put in the other half. | Both halves are denied: the target on its own, and on Linux the directory holding the link WHOLE, which is the only handle on it. A link whose target is not there yet is denied where it lands, so it cannot be created. A chain that goes through further links is held hop by hop, and so is the path a `gitdir:` pointer's or a `commondir`'s value walks, which git resolves the same way. See **An entry that is itself a symlink**. |

## How a pointer and a `commondir` are read

- **Byte for byte, as git's `read_gitfile_gently()` does.** A differential test against real git is what holds this (see Testing).
- **A target that is not a git directory is not followed. One that does not exist yet is denied anyway,** so the command cannot create it and fill it with hooks first.
- **`..` after a symlink:** `link/../x` does not land where folding the string on paper says, so both the lexical answer and the kernel's are computed, and where they differ both directories are denied.
- **A target this process cannot read is denied whole,** at its deepest reachable ancestor. git, running as the same user, cannot read through it either, so that repository is already broken for everyone; refusing instead would be a brick one `chmod` plants.
- **Where git's answer cannot be known at all, the wrap is refused** - see **Refusals a command can plant**.

## Failing closed

ripgrep lists the rest of the tree and exits non-zero when an entry gets in its way, so a failed run still carries matches. Whether they can be used is decided by the `(os error N)` codes on its stderr, never by the paths printed beside them: a directory's name is chosen by whoever created it and reaches stderr unescaped, so a name holding `: `, a trailing space or a newline would otherwise pick the path that gets denied. A name can only add codes, and an unrecognised code refuses.

- every code `ENOENT` - entries that went away while the scan walked. There is nothing at those paths to deny, and mounting one would plant a file on the host exactly where the command deleted one. The listing still counts.
- any code `EACCES` - a directory it could not read holds an unknown tree, so it is denied whole and the listing still counts. WHICH directories is settled by a bounded walk of the working directory, to the scan's own depth, skipping `node_modules` as the scan does, and inside what is left of the scan's own time budget. A directory that is readable but not SEARCHABLE (mode 0400) lists its entries and lets nothing through them, so it is that directory - not the child whose own listing failed for want of its search bit - that is denied: naming the child asks bubblewrap for a file mount at what is really a directory, and fails every later wrap in the tree with an untyped error. A directory at the very depth the scan stops at is a case of its own: ripgrep's walker opens one to list it and reports it when it cannot (reported for versions up to at least 15.0.0; 15.2.0, which is what I measured on, does not), while this walk stopped a level above - so such an EACCES had nothing found for it and the wrap was refused, over a layout as ordinary as a `docker/data/postgres` another user owns. Nothing below such a directory is in the scan's reach on either side, so it is now accounted for with nothing to deny.
- anything else, no code at all, or an `EACCES` the walk finds no directory for: the command is refused with `deny_scan_failed`, and the raw stderr (truncated) is in the message.

Those three rules read a run that exited on its own. A run that was KILLED is refused whatever its stderr held, because its listing stops wherever the kill arrived and nothing on stderr says what lay past that: the timeout, an OOM kill, a signal to the process group. Until this round only the timeout refused, and the others fell through to the codes - so a run killed mid-listing inside its budget, whose stderr already carried the one `(os error 13)` line a `chmod 000` produces, had every code it printed stood in for and its partial listing was used as a complete one. The message names the signal and says whether the run had used the time it was given. A scan that could not be run at all - the binary could not be spawned, which is the dependency the start-up check refuses on, met later - is refused the same way: the command about to run is the one that could have made the tree slow to walk. An abort is the caller's own answer and comes back as the `AbortError` it is.

The same whole-directory deny stands in elsewhere: for a directory the `.git/modules` walk could not list, and for an entry whose target is there and cannot be inspected. What is denied is the deepest directory that can be reached towards it - towards the TARGET for an entry that is a symlink, rather than the directory holding it, which would be every submodule beside it - and it may be `.git/modules` itself where the walk cannot get past it. An entry of a `.git/modules` that is a symlink loop reaches no directory at all, so nothing is denied for it: there is nothing behind it to protect, and a bind cannot be put at the link's own location either, since mounting there means resolving it. A git directory's OWN entry that loops is a different matter, and a chain that reaches nothing is held by the links it went through - see **An entry that is itself a symlink**.

The `.git/modules` walk is bounded the same way, by as much time as the scan gets. It has no depth bound of its own: each directory is visited once, keyed by where it really is, so a symlinked entry pointing back into the tree ends that branch, and what bounds the walk is time - which is the answer a bind mount pointed at its own ancestor needs, since that gives every level a real path of its own and has no end the visited set can see. Running out of it is `deny_scan_failed` like the scan's own overrun: a walk that stops somewhere unknown would leave the submodule git directories it never reached with writable hooks. The walk keeps its own stack, so a chain two hundred directories deep cannot overflow this process's. It looks at the clock as the entries go by rather than once per directory, so one directory a command filled with fifty thousand entries is stopped in the middle rather than at the end of it, and each directory is listed once. A walk made after the scan takes its own budget of the same length: the scan has a timeout of its own, and what a slow but successful one left would otherwise refuse a repository whose submodules a walk given the time gets through in microseconds.

A git directory that takes no mount point at all - one this process cannot write, such as a root-owned vendored checkout or a read-only mount - is bound read-only whole rather than refusing. It needs nothing written to the host, costs what was already unwritable, and is what stops the command from making the directory writable and creating the redirect file itself. `deny_placeholder_unavailable` is left for the cases where the host genuinely cannot take the file: no space, or no writable temporary directory for the placeholder.

The one consumer that must not be refused is the violation monitor, started once for a session to judge a refused write against the same cwd denies. It asks through a wrapper that falls back to this directory's plain deny paths and warns, so one unreadable repository does not fail the whole start-up; every wrap in it is still refused.

## Refusals a command can plant

These wrap-time refusals follow from what a sandboxed command is allowed to write, and the command that would put it right is refused as well, because every wrap in that working directory hits the same thing. **Each is lifted only by a command run outside the sandbox**, and each names the offending path, or the line the scan printed, in the message so that the operator knows which:

- `deny_git_metadata_unreadable` - a `.git` pointer or a `commondir` whose target cannot be worked out the way git works it out: bytes that are not valid UTF-8, a `commondir` larger than the size read here, or one that is there and cannot be read. Creating a new `.git` pointer inside an allowed write path is deliberately allowed, so `printf 'gitdir: \xff' > sub/.git` is one command. On macOS the same condition is a `GitMetadataError`, which now carries a name and a `.code` and is exported from the package root.
- `deny_unresolvable_git_entry` - an entry of the working directory's OWN repository (`hooks`, `config`, `commondir` or `config.worktree` of the git directory its `.git` is or names) is a symlink whose chain reaches NOTHING (a loop, or a name no file can occupy) and one of the links it goes through sits in the working directory or a path this command may write, where no read-only bind can hold it. There is no landing to deny instead, so the command is refused rather than run with the link free. The message names the entry, the link and the directory holding it; repairing the chain, moving the links inside the git directory, or a `denyWrite` entry naming the directory holding them, each lifts it. This is the layout `main` already refuses, by a different route (the deny path cannot be resolved, so the wrap fails closed); a pointer's value that reaches nothing is deliberately NOT this - see below. Neither is the same chain behind anything a command may have made: a nested repository the scan found, a submodule's git directory, a `modules` that was not there. Those get the warning an unheld hop gets and the command runs, because refusing there let a few ordinary writes lock every later command out and protected nothing the command could not already reach by creating a nested repository outright. The four entries of the working directory's own repository are denied at their own paths from the first wrap that saw it, so a link found at one is the caller's layout; what is left to plant is a repository created and linked this way by one command in a directory that had none.
- `deny_scan_failed` for a scan failure no deny stands in for, as above, and for a `.git/modules` tree that takes longer to walk than the scan's own budget - a tree wide or slow enough is one a command can make, and removing what makes it slow lifts it. On macOS the same condition is a `SubmoduleWalkBudgetError`, exported from the package root.
- `too_many_arguments` where a command plants enough CHECKED-OUT submodules - a `.git` pointer file and the git directory it names - that the profile does not fit with every submodule deny degraded: about 1475 of them, measured below.

**No whole-directory deny is ever emitted for the work tree, whichever producer asked for it.** Every one of them is a stand-in read off paths a sandboxed command can write, so each is a way to aim one at the project: a `.git` file naming the checkout (`gitdir: .`, with one entry of the checkout a symlink, makes the checkout a git directory with a symlinked entry - denied whole), a `.git/modules` entry linked at `../..`, a git directory entry pointing at the checkout, a `chmod` that leaves the working directory unreadable. Any of them bound the working directory or a write root read-only from the next wrap on, and nothing inside the sandbox could undo it. Now: a directory that can be LISTED is dropped with one warning, since what that deny was holding keeps the deny paths of its own two ends; a directory that cannot be listed stands in for a tree nothing else covers, so the command is REFUSED (`deny_scan_failed`, naming the directory) rather than run against an unscanned tree or locked out of its own project - one `chmod` outside the sandbox lifts it. The same call is made on macOS, where a literal entry is a subpath deny.

Everything else a command can do to the tree is answered with a deny rather than a refusal: an unreadable directory is denied whole, an unwritable git directory is denied whole, and a vanished file is nothing. A wide `.git/modules` is one of those: it is writable inside the sandbox and a command can fill it, and what that costs the commands after it is degraded denies rather than a refusal - git writes inside the submodules that were degraded fail read-only until the entries are removed, which takes a command run outside the sandbox, since what holds them is by then read-only inside it.

A held chain hop belongs in that same family, and it is worth saying plainly rather than presenting the holder bind only as the cost of a layout the caller already had: the layout need not pre-exist. A command can make one - `mkdir -p x/.git; touch x/.git/HEAD; ln -s ../../src/hop x/.git/hooks; ln -s /nowhere src/hop` - and from the next wrap on `src/` is bound read-only for holding the link, which cannot be removed from inside it either. That is a denial of service by code that could as easily have deleted `src/`, and it is the reason to hold the hop rather than a reason not to.

## Finding a nested repository

A nested repository is recognised by any regular file directly inside its `.git`, which is why it is found at the depth the repository sits at rather than one level further down where the hook files are. With `allowGitConfig` off - the default - the deny of `config` keeps such a file there and it cannot be removed from inside the sandbox, so one command cannot empty the directory and hide the repository from the next command's scan. With `allowGitConfig` on it could, so each command additionally walks the working directory to the same depth (the same walk the `EACCES` case uses) and denies the `.git` directories it finds. That walk is the cost of allowing config writes and is not run otherwise.

## Submodules past the argument cap

bubblewrap parses at most 9000 words and takes three of them for each mount. A submodule git directory costs four denies and an ancestor pin; a CHECKED-OUT one costs its `.git` pointer file and the pin of the directory holding it on top of that. A repository with a few hundred submodules therefore asks for more mounts than the profile can carry, and on `main` every command in it is refused up front with `too_many_arguments`. That ceiling is bubblewrap's own and not this library's to lift - but refusing the repository is the wrong answer to it, because a repository with thousands of submodules then gets no sandbox at all.

So the denies degrade instead, fail-closed and only as far as they have to. The wrap builds the profile, counts the words bubblewrap will actually parse, and while that is past the cap gives up one step and builds again:

1. precise denies (`hooks`, `config`, `commondir`, `config.worktree`) while the profile fits;
2. a submodule git directory that does not fit is denied WHOLE, one read-only bind. git writes inside that submodule fail read-only, its `objects`, `refs` and `index` with them; nothing outside it changes. A checked-out submodule's `.git` pointer names the same git directory, so its copy of those denies goes with them, and the pointer file keeps its own deny - it names a file, and a file has nothing to collapse into. So does what an entry that is a symlink leads to: it lies outside the git directory and the bind does not cover it, so it costs a mount the degrade does not give back, and where the budget cannot take that the degrade goes one step further rather than dropping it;
3. where not even one bind each fits, the enclosing `modules` directory is denied whole and every submodule really under it goes with it. An entry the walk reached through a symlink can have its real path anywhere, and one whose real path is outside keeps the deny it had: a bind over `modules` does not cover it.

What is given up is measured rather than predicted, which is the point: the ancestor pins a deny needs are known only once the deny list is complete, so what a repository's submodules really cost cannot be worked out before the profile is built. A count of the deny paths inside the write allowlist says where the passes start, so a repository with ten thousand submodules is not built out in full first; it is a floor, and the measurement is what settles on a level.

Which submodules keep their precise denies is settled by sorted order, the last given up first, and the working directory's own repository is served before the nested ones the scan found, which are sorted before they are served - ripgrep lists a tree in whatever order its threads finish it in. So two wraps of one tree produce the same argv and the second command in it is sandboxed exactly like the first. One `SRT_DEBUG` warning per wrap names what was given up, at which level, and for which repositories; there is no other wrap-time channel to report it on.

A whole-directory deny has fewer destinations to seed ancestor pins from, and the one it does have is the git directory itself, so renaming `.git/modules/<name>` out from under the deny still fails `EBUSY`.

The violation monitor works the same denies out for itself, through the same function over the same inputs, with nothing degraded. It is computed once for the session, so a command that forces a degrade later leaves it naming the precise path inside a submodule git directory the wrap has by then denied whole: the write is refused either way, and it is the path in the report that is stale, never what bubblewrap enforces.

macOS is unchanged: a Seatbelt profile is a text file with no cap on how many entries it carries, so nothing is ever degraded there.

## An entry that is itself a symlink

`hooks`, `config`, `commondir`, `config.worktree` and an entry of a `.git/modules` can each be a symlink, and then there are two things to protect, not one: what a write through the link reaches, and the link itself - which a command can unlink and leave its own `hooks/` in place of, for the host's git to run before the next command's scan sees anything. Which of the two a single deny covers is the backend's, and the two backends differ:

- **Linux** resolves a deny path before it mounts it, so the bind lands on the target and the link's own path carries nothing. Measured under bubblewrap 0.11: a write through the link is refused, `rm` and `mv` of the link succeed.
- **macOS** compares its filter against the path the kernel resolved the operation to, while the path IN the filter is compared as a string. For an unlink or a rename that resolved path is the link's own, since the last component is not followed; for a write through the link it is the target. Measured under `sandbox-exec`: with `(deny file-write* (subpath <link>))` the `rm` and the `mv` of the link are refused and the write through it lands; with the target denied as well, both are refused.

So both halves are named. The target gets a deny of its own, which is what macOS was missing. And on Linux the directory holding the link is denied WHOLE - the git directory for one of its own entries, the directory holding a `.git/modules` entry for one of those - since a mount cannot be put on a link's own path: putting one there means resolving it. A git directory a `gitdir:` pointer leads to is one of them too: a linked worktree's main repository is reached that way and nothing else holds it. macOS needs no whole-directory deny and does not get one, so git keeps working there.

A chain can be longer than one link, and then two halves are not the whole of it. An entry can go through a link on the way - `hooks -> ../hooks-link` with `hooks-link -> .githooks`, or `hooks -> ../shared/hooks` with `shared` a symlinked directory - and so can the VALUE of a `gitdir:` pointer or a `commondir`, which is a path git walks as the kernel does: `gitdir: hop/../evil`, or a git directory reached through a symlinked component. Either way the file's or the entry's deny holds that end, the landing's deny holds the other, and a link BETWEEN them is held by neither. A command that retargets it, or renames it aside and puts its own directory there, moves where the chain resolves without touching either end, and the host's git follows the chain afterwards. So the walk that resolves an entry, and the one that resolves a pointer's value, now hand back every hop, and each hop is held the way the entry or the pointer file is: its own path is a deny path, which is what holds it where a filter matches the name a rename or an unlink uses, and on Linux the directory HOLDING it is denied whole, which is what holds it where a deny resolves. A hop inside the git directory, or inside the directory holding a `.git/modules` entry, needs no directory of its own - the whole deny that entry already brings covers it - and one no allowed write path contains is read-only anyway. Linux drops the hop's own path instead of binding it: a bind there lands on whatever the link leads to, which for a symlinked directory component is a whole directory nothing asked to deny. A chain past the kernel's own hop limit reaches nothing - git fails on it too - so it has no hop to hold, and the whole-directory deny its landing gets is the whole answer. A link ABOVE the entry or the pointer file is not a hop of the chain at all: a checkout reached through a symlink is where the repository lives, and each walk starts where the kernel already is.

One thing more is needed on macOS, and it is what this round adds. Holding each hop by its own name is only half of a spelling rule the two backends answer oppositely, and the measurement above generalises: a filter whose path goes THROUGH a link is compared against a resolved path it can never equal, so it matches nothing at all. Three spellings of one object, under `(deny file-write* (subpath P))` with `dlink` leading out of the tree:

| P | write `P/f` | write through `dlink` | `rm`/`mv dlink` |
| --- | --- | --- | --- |
| `dlink/sub` (through the link) | lands | lands | lands |
| `dlink` (the link itself) | lands | lands | REFUSED |
| the landing | REFUSED | REFUSED | lands |

`literal` and `regex` filters answer the same, and so does `file-read*`, where a deny on the link additionally stops the traversal through it. So the path as written is the only name an unlink or a rename of a link uses, the landing is the only one a write to what it leads to matches, and a chain needs both. Until now the macOS entries carried the landing only where a `..` made the two PHYSICAL resolutions differ; `gitdir: links/shared/gd`, with no `..` in it, left every deny under the git directory it names spelled through `links` and covering nothing. Every path the git producers read off disk is now emitted twice there, as written and at the landing, de-duplicated - which for a repository with no link in any of those paths is the same string and the same single entry. Linux needs none of it: bubblewrap resolves a deny path before it binds and lands on the same directory whichever spelling it was handed, and the 84-configuration argv dump is byte-identical.

One kind of deny path takes its own spelling and only its own: a HOP. What a hop leads to is a whole directory nothing asked to deny - the `/private/var` a `/var/folders` pointer value walks through, the shared directory a `links/gd` value crosses - and denying it takes the checkout and everything beside it read-only, which is how macOS CI caught the first cut of this: two linked-worktree cases whose positive control, an ordinary write in the checkout, stopped succeeding. The link's own name is the whole handle on a hop, which is the mirror image of Linux dropping hop links from its binds for the same reason. Nothing of the fix is given up by it: what lies past a hop is the git directory, whose own deny paths are in the same list and do carry both spellings.

The other prefixes this could touch are already right. `/tmp` and `/var` are ordinary symlinks, and a filter spelled with either matches nothing - measured - but path normalisation already rewrites both onto `/private`, which is the spelling Seatbelt compares against. `/Users` is a firmlink rather than a symlink and canonicalises the other way, the `/Users` spelling being the one Seatbelt uses; `getcwd` returns it, so the cwd every mandatory deny is joined onto is already the right spelling and already free of symlinks (measured: `os.getcwd()` from a directory reached through a link returns the physical path, and from `/System/Volumes/Data/Users/<user>` returns `/Users/<user>`).

That a pointer's value is walked this way is git's rule and not a guess, checked against real git 2.55 and pinned by two cases in the parity test: with the kernel's landing and the lexical fold both git directories, git opens the kernel's landing; with only the lexical one there, git refuses the checkout outright (`not a git repository`). And the hazard itself is measured, not reasoned: `ln -sfn` of one link between two runs moves `git rev-parse --absolute-git-dir` from one git directory to another, with the pointer file and both directories untouched.

- **A link whose target is not there yet** is not passed over: where it lands is denied as a git directory that does not exist, which is what blocks a command from creating it and filling it with hooks.
- **`modules` is held as a link, and is no deny path of its own.** A git directory keeps its submodules' git directories under `modules`, each with precise denies of its own, so denying that directory whole would take every one of them read-only. It is an entry like the others in the one respect that matters: remove the link, put a directory of your own where it was, and every submodule git directory the repository keeps is one nothing denied - `rm .git/modules/foo/modules && mkdir -p .git/modules/foo/modules/bar/hooks && echo ... > .../post-checkout` left the hook on the host. So it is held the way a chain hop is: its own name where a filter matches the name a rename uses, and the directory holding it - that git directory, denied whole - where a deny resolves. The top-level `.git/modules` is held the same way; only a `.git` that is itself a link remains open, since there the only handle above it is the work tree root. **What it costs, and that a command can arrange it:** an ordinary repository has no `.git/modules`, and an absent one is deliberately not denied (a mount there would stop `git submodule add`), so one `ln -s` inside the sandbox makes `.git` a git directory with a symlinked entry - and from the next wrap on `git add`, `git commit` and `git checkout -b` fail read-only in that repository, with nothing inside the sandbox able to remove the link. `git status`, `log` and `diff` keep working, and removing the link from outside the sandbox puts it back. It is the same family as the other costs a command can plant, and it is the price of holding the link at all.
- **An entry that cannot be classified at all** - an `lstat` that fails for anything but absence - is taken for a link, both denies and all.
- **A chain that reaches NOTHING is held by its hops alone.** A loop, or a chain longer than the kernel follows, has no landing: git gives up on it exactly as this walk does, so the deny the landing would get stands on nothing. What is left is every link the walk visited before it gave up - put a directory in one of their places and the chain reaches it, and the host's git reads the hooks or config inside. That is measured against real git rather than reasoned: with `.git/hooks -> ../links/a` and `a -> b -> a`, git commits with no hook; replace `links/a` with a directory holding a `pre-commit`, entry and git directory untouched, and git runs it. So each such link is denied like any other hop, named once however many times the loop met it. A loop confined to the git directory needs nothing more - the whole-directory deny a symlinked entry already brings covers its links. One that runs out through the work tree is held by the directories holding those links; where THAT cannot be done, because such a directory is the working directory or a write root, Linux refuses the command with `deny_unresolvable_git_entry`, since the warning that stands in for a bind elsewhere leans on both ends of the chain being denied and here there is no far end. Before this round the loop's links were left free and the command ran, which was looser than `main`, and silently so.
- **A pointer VALUE that reaches nothing is not denied whole anywhere.** Neither the directory it folds to on paper nor the deepest one still reachable towards it, which for `gitdir: a/gd` with `a` a loop in the checkout is the checkout itself - denying that whole made every write in the project fail. A `.git` file is something a sandboxed command is deliberately allowed to create, so failing closed there let one command lock every later command out of the project, and refusing instead would be the same lockout by another name. git opens nothing through such a value (pinned against real git), so what is denied is the pointer file and the links the walk did go through, with one warning naming the file. What a command could reach by re-pointing a link it is free to write, it could reach by creating a nested repository outright, which is a recorded limit of scanning before a command rather than after it.
- **A hop of the chain that cannot be read** leaves the rest of it unknown, so the git directory is denied whole as it is for an entry that cannot be classified, and every hop the walk did read is held all the same.
- **What it costs** is git writes inside that one repository: `git status`, `log` and `diff` keep working, while `git add`, `git commit` and `git checkout -b` fail read-only until the link is gone. A symlink planted under a `.git/modules` costs the same way, for every submodule under it, from the next command on - one more command-plantable degrade, never a refusal and never a widening.

## Who sees a difference

- **An ordinary repository** keeps its index, objects, refs and `git commit`, which a test holds by committing inside the sandbox twice over the same repository. It does see one change on Linux, and on the host as much as in the sandbox: for as long as a wrapped command runs, its git directory holds a `commondir` file containing `.` (and so does each submodule's), written by the wrap before bubblewrap starts and removed after the command. `.` is what git concludes with no `commondir` there, so every git command keeps working - but the file's existence is not nothing: `git rev-parse --git-common-dir` and `--git-path` print the absolute real path where they printed a relative one. A process killed before its cleanup leaves the file behind; git reads it happily, and the next wrap in that repository takes it away. An empty `commondir`, which git refuses outright and which is what older releases left, is repaired the same way - unless it still has write bits, which means a host `git` is between creating the file and writing it, and then it is left exactly as found and only covered for the command. The exception is a tree that trips a whole-directory deny: it is read-only inside the sandbox, `objects`, `refs` and `index` included. Scan depth and Windows are unchanged. Not behind a flag.
- **A command whose scan fails** is refused rather than wrapped with the working directory's own denies alone: the scan is where every nested repository below it comes from, so a command that blinds it - one `chmod` of one directory was enough - would otherwise leave the next command a tree of writable hooks. The failures that still wrap are the two above, and neither takes a path from what the run printed.
- **A repository whose `hooks`, `config`, `commondir` or `config.worktree` is a symlink** - a shared hooks directory linked into `.git`, say - has its git directory denied whole on Linux: `git status`, `log` and `diff` keep working, `git add`, `git commit` and `git checkout -b` fail read-only, and removing the link puts it back. Nothing changes for it on macOS. Measured in a plain repository on the same box, `git commit` inside the sandbox already fails for a reason of its own (the loose-object temporary file), so what the whole deny adds on top is the index writes.
- **A repository whose entry reaches its target through a second link** has the directory HOLDING that link denied whole on Linux as well: with `.git/hooks -> ../links/hop`, `links` is read-only for the command, and only that directory - never what the link leads to. Where the directory holding the hop is the working directory or a write root there is nothing to bind over it, so the command is not refused and one `SRT_DEBUG` warning per wrap names the entry and the hop instead; see Known limits. macOS gains the hop's own path as a deny and still denies nothing whole. The argv dump over 84 configurations differs in exactly the sections with a hop a subdirectory holds, and the submodule cost measurements are identical at every size.
- **A checkout whose `.git` pointer names its git directory through a link** is the same shape from the other end, and is held the same way: with `gitdir: links/hop/gd`, `links` is read-only for the command on Linux and the link's own path is a deny entry on macOS. In the common case the link sits in the checkout itself, which is the working directory, so what the wrap can do is the warning - now naming the pointer file and the directory it leads to rather than an entry. Of the six new dump sections, three gain exactly one read-only bind (the directory holding the hop), two are identical (the working directory holds the hop, so it is warned about; and a hop outside every write root, which is read-only already), and one - a value with a `..` after the link - gains that bind and loses the stub it used to plant at the path's lexical fold, which now sits inside a directory denied whole.
- **A repository whose git directory entry runs into a symlink LOOP** is the one place this branch had been looser than `main`, and is not any more. `main` refuses such a command outright (the deny path cannot be resolved, so the wrap fails closed); this branch used to run it with the loop's own links free, and no warning, so a command could put a directory in one of their places and the host's git would read the hooks inside. Now: the links are held by the directory holding them where one can be bound (one extra read-only bind, and the project stays writable), and where it cannot - the working directory or a write root - the command is refused with a reason of its own. Same outcome as `main` for that layout, different message.
- **A checkout whose `.git` pointer value runs into a loop** is the opposite change, in the other direction. It used to fail closed on the deepest directory still reachable towards the value, which for a loop in the checkout is the checkout: the whole project went read-only, for a file a sandboxed command may create, which made it a denial of service plantable from inside the sandbox in one command. Now the pointer file and the loop's links are denied and nothing else, with one warning naming the file. A command after the plant writes the project as before; what it cannot do is rewrite or remove that `.git` file, which takes a command run outside the sandbox.
- **A command whose scan was killed mid-listing** is refused (`deny_scan_failed`) whatever its stderr held, where a kill other than the timeout used to fall through to the error codes and could leave a partial listing treated as complete.
- **A macOS repository whose git metadata is reached through a symlink** gains the denies it was missing: a hook written into the git directory a `gitdir:` pointer names through a link, or into a submodule behind a symlinked `.git/modules` entry, was landing under both spellings and is now refused under both, and a `commondir` or `config.worktree` cannot be created at the landing. Run under real `sandbox-exec` with profiles this branch's generator produced, the three writes that landed at the previous head are refused at this one, while the link's own name stays held and an ordinary project write still succeeds. What it costs: nothing at all for a repository with no such link - the entry count, the rule count and the profile are byte-identical at 0, 50 and 700 submodules - and for one where every `.git/modules` entry IS a symlink, entries go 238 to 438 at 50 and 2838 to 5638 at 700, the profile 31 KB to 50 KB, and `sandbox-exec` start-up 43 ms to 70 ms at that size. The SBPL rule count does not move at all (31 either way): the extra paths are filters inside rules that were there already. Building the entry list costs the walk of each path: 32 ms to 77 ms at 700 submodules, unchanged below 50.
- **The degrade warning** no longer counts what a whole-directory deny of the same plan had already taken. A `.git/modules` denied whole for a symlinked entry leaves every submodule under it read-only before anything is degraded, so degrading those takes nothing away, and the warning now describes what is actually read-only rather than reporting a cost the profile never paid.
- **A caller reading `.code`** gets two new `LinuxSandboxProfileError` codes, `deny_git_metadata_unreadable` and `deny_unresolvable_git_entry`, and `GitMetadataError` from the package root on macOS. `RipgrepConfig.timeoutMs` is now in `RipgrepConfigSchema`, where the type already had it.
- **Git operations these denies break** (listed in the README): `rm -rf` or `git clean -ffdx` over a submodule checkout or linked worktree; `git worktree remove`/`move`/`repair`; `git submodule deinit`; `git submodule update --init` for a submodule not yet cloned; `git push -u` and `git checkout -b x origin/y` from a linked-worktree cwd; `git init`/`git clone` into a subdirectory.
- **Linux command-line cost:** a git directory now costs up to four read-only binds instead of two, each submodule is one git directory, #514 pins each of those directories on top of that, and a wrap that mounts a redirect placeholder adds one more bind for the store it comes from. The 128 KiB `execve` limit that put the ceiling near 240 submodules, past which every sandboxed command failed with `E2BIG`, no longer applies: #504 has moved the mounts off the command line, leaving bubblewrap's own limit of 9000 arguments. Measured on this head, with one write root and no network restrictions: 580 submodule git directories still keep every precise deny and 600 give up 8 of them; 2000 keep 242 precise; a repository with 10000 goes behind one bind of `.git/modules`. With a `.git` pointer file each, 420 keep every precise deny, 1470 still wrap, and 1490 are refused with `too_many_arguments`. Building the profile takes 14 ms with no submodules and 180 ms at 500, where nothing is given up; a repository that degrades builds more than once, which is 380 ms at 600, 710 ms at 2000 and 310 ms at 10000, per command.
- **A repository whose `modules` directory is a symlink** has its git directory denied whole, as one whose `hooks` is: `git status`, `log` and `diff` keep working, `git add`, `git commit` and `git checkout -b` fail read-only, and removing the link puts it back. That is a repository a command can make out of an ordinary one with a single `ln -s`, which is stated above and in the README rather than left to be found.
- **A project with a directory that is readable but not searchable** (mode 0400, unprivileged only) keeps wrapping. It used to break every later command with `bwrap: Can't create file at <path>: Is a directory`, because the child whose listing failed - a path the host cannot stat either - was recorded instead of the directory that cannot be seen through; only a `chmod` outside the sandbox cleared it.
- **A project with a directory another user owns exactly at the scan's depth limit** (`docker/data/postgres`, mode 0700) keeps wrapping too, where ripgrep reported the EACCES and this walk, stopping a level above, found nothing to stand in for it and refused every command.
- **A repository that keeps linked worktrees** has `commondir`, `gitdir` and `config.worktree` of each `.git/worktrees/<id>/` denied, so `git worktree prune` (and the prune `git gc` runs) cannot remove a stale entry from inside the sandbox: it prints one `failed to delete` line, exits 0 all the same and leaves the entry for the next prune outside, and a detached `git gc --auto` leaves no `gc.log` over it (measured with git 2.55). Nothing else changes for it.
- **A tree holding a git directory this process cannot write** (mode 0555, or another user's) keeps wrapping where every command used to die at bubblewrap start-up.
- **A repository with hundreds of submodules outside every write root** wraps in a few builds rather than one per submodule.
- **A caller of the Linux violation monitor** gets the cwd's denies with nothing degraded, computed once for the session as before. Where a wrap degrades a repository mid-session the monitor names the precise path inside a submodule git directory the wrap has by then denied whole; the write is refused either way.

## Known limits

A pointer file or a pattern-matched path is protected where it is, so a command could rename the directory *holding* it and put a fresh one in its place (`mv lib lib.old && mkdir lib && echo 'gitdir: ...' > lib/.git`). On macOS that is still open for the pattern denies; the literal ones are safe from it. On Linux it is closed for everything the scan reached, now that #514 is in: the ancestors of every denied path are pinned, and on the merged tree under bubblewrap `mv` of the directory holding a denied pointer file, and of the package directory above a nested repository's hooks, both fail with `EBUSY`.

The scan never looks inside a `node_modules`, so a repository under one is never found and its hooks stay writable. That exclusion is permanent: the scan walks gitignored data as it is, and a `node_modules` is where the cost of that would land.

A tool that reads the mere existence of a `commondir` as "this is a linked worktree" misreads an ordinary repository while a wrapped command runs there, and until the next wrap if a killed process left the file. No content avoids that - the alternative is a file git cannot read.

A `commondir` holding exactly `.` is recognised as this wrapper's own by its content alone, so one a user put there by hand is removed by the next wrap's cleanup along with the wrapper's.

Cleanup deferral is per process. Two processes wrapping commands in one repository can each claim the same mount point, and the one that cleans up first removes it while the other's sandbox is still running, which detaches that bind and stops that one deny applying inside it. Binding the placeholder from the store rather than from the mount point keeps the sandbox starting; the deny is what is lost.

The `.git/modules` walk follows a symlinked entry, because git does. That has two consequences, neither of them an escalation - each is a command making something read-only for the commands that follow it, never writable.

<details><summary>The two symlink consequences</summary>

- A symlinked entry is denied under the path as written, which on Linux the backend resolves to the link's target, and the directory holding the link is now denied whole for the link's own sake. A command that can write under `.git/modules` can therefore point one directory of its choosing at itself and have its hooks and config mounted read-only for the next command, and leave that `.git/modules` read-only as well until the link is removed.
- A symlinked entry whose target is there and cannot be inspected is denied at the deepest directory that can be reached towards THAT target, so it is that target's tree - not everything under `.git/modules` - that goes read-only; the `.git/modules` itself is denied for the link, as above.

</details>

**A link that IS the git directory.** The denies above hold a link INSIDE a git directory or a `.git/modules`. A `.git` that is itself a symlink to a git directory elsewhere is not one of them (a `.git/modules` that is a link IS held now, by the git directory holding it): what it points at is denied as any git directory is, and the link itself sits in the working tree, which the sandbox exists to let the command write. Replacing it makes the repository a different one, which the next command's scan covers - but a host git run between the two commands follows the new link. There is nothing to bind over: a mount at the link's own path resolves it, and the only handle above it is the working tree root.

**A hop the working directory or a write root holds.** `.git/hooks -> ../hooks-link`, with `hooks-link` in the repository root, is the common shape of a chain with a hop in it, and the directory holding that hop is the working directory itself. A `.git` pointer whose value walks through a link in its own checkout is the same shape by construction, since the checkout IS the working directory. A read-only bind of it would take the whole tree the sandbox exists to let the command write, so the command is not refused for it either - as long as the chain reaches something, which is what leaves both its ends denied. A chain that reaches nothing through such a hop has no far end to deny and is refused (`deny_unresolvable_git_entry`), unless it is a pointer's value, where the refusal would be the lockout it exists to avoid. One `SRT_DEBUG` warning per wrap names the hop and what leads through it - the entry, or the pointer file and the git directory its value names - and says it can be retargeted from inside the sandbox; both ends of the chain stay denied, and the next command's wrap sees the new chain. A `denyWrite` entry naming the directory holding the link closes it at the cost of that directory; moving the link inside the git directory, or spelling the pointer's path without it, closes it outright. Snapshotting each such hop and restoring it after the command was tried and not adopted: measured under bubblewrap, a hop the command replaced with a real directory cannot be put back by a symlink call (`EEXIST`), so a restore means removing what a command wrote inside the caller's own work tree, which this library never does - it only removes mount points it made itself. `cleanupAfterCommand` is also the embedder's to call, and nothing of it closes the window while the command runs, which is the window a host git run between two commands falls into.

**A CALLER's own deny spelled through a symlinked component (macOS).** The rule above is closed for every path the git producers read off disk. It is not closed for a `denyWrite` or `denyRead` entry the caller wrote, and that is pre-existing rather than introduced here. Such an entry is normalised by resolving it, which is the right spelling for Seatbelt - but the resolution is accepted only when it stays inside the entry's own tree (`isSymlinkOutsideBoundary`), so a deny naming a link that leads ELSEWHERE, or naming a path through one, keeps the spelling as written: on macOS it then holds the link's own name and nothing under it, while on Linux the bind lands on the target as always. An entry naming a path that is not there yet keeps its spelling for the same reason, realpath having nothing to resolve. Emitting both spellings there is the same fix, in a place where it reaches every caller's config rather than paths this library computed, so it wants its own measurement of what doubling a deny costs a profile and its own pass over the carve-out and re-emit rules that read those entries back. Left for a follow-up, deliberately, and named here rather than fixed quietly. The same applies to a DANGEROUS file name that is itself a symlink - a `.env` in the working directory linked to a file outside it - whose deny holds the name and not what it leads to; the working directory the name is joined onto is not affected, `getcwd` being physical.

**A loop under `.git/modules`.** An entry of a `.git/modules` whose chain reaches nothing is left exactly as it was: nothing denied, one warning. The links it goes through are not held either, and that is deliberate rather than an oversight - a `.git/modules` is writable inside the sandbox, so putting a directory at one of them to make the entry reach it is no shorter a way to a planted submodule git directory than creating one outright, and the only handle on those links that this backend has is a whole-directory deny of the `.git/modules`, which would take every submodule beside them read-only over one link a command can write. A git directory's own entries are the other case, where the whole-directory deny already holds the entry and the hops are all that is left open.

**A hook FILE that is a symlink into the work tree.** The install idiom `ln -s ../../scripts/pre-commit .git/hooks/pre-commit`, with `hooks` itself a real directory, is not one of the entries above. The hooks directory is denied, so no name can be added there and none re-linked (measured: `ln -sfn` over the hook name fails "Read-only file system"), but the script the link leads to is ordinary project content and its CONTENT is writable, directly and through the link, on every build. That is the sandbox doing its job: a tracked script in the work tree is exactly what a command is there to edit, and nothing distinguishes the one the user chose to run as a hook from any other file in the repository. A `denyWrite` entry naming the script is the remedy where it matters.

**A pre-existing HARD link of a denied file.** A bind protects a path, not an inode, so an append through a second name for `.git/config` elsewhere in the work tree lands. A sandboxed command cannot make such a link to a bound file - the bind is its own mount and `link(2)` across one fails `EXDEV`, measured as "Invalid cross-device link" - so it has to pre-exist.

Neither the walk's depth nor its breadth is bounded by a number; what bounds both is time, as much as the scan gets, and a very wide or very deep `.git/modules` costs one `readdir` per directory until it runs out.

Past about 590 submodule git directories - about 425 where each is checked out - the precise denies stop fitting and the profile starts giving them up, and past about 2950 the whole `.git/modules` goes behind one bind. What that costs is git writes inside those submodules: `objects`, `refs` and `index` are read-only there for as long as the command runs, so a `git commit` inside one fails where it used to work, and clearing the tree that caused it takes a command run outside the sandbox. The exact figures depend on what else the configuration mounts, which is why the wrap counts rather than assumes, and a warning names what was given up.

`too_many_arguments` is what is left for a profile that does not fit with every submodule deny given up. Measured: about 1475 CHECKED-OUT submodules, whose pointer files and the pins under them no degrading saves, and a few hundred separate nested repositories, whose own four denies are not submodule denies. Bare submodule git directories do not reach it at any count, multi-segment (`vendor/lib`) and nested names included: the one bind over `.git/modules` is where they end.

**A `core.hooksPath` or `include.path` into the work tree.** The denied `config` is never read. Where the repository's own config already sets `core.hooksPath` to a directory in the work tree (husky's `.husky/_`), git never looks in `.git/hooks`, and what it runs is ordinary writable content; the same goes for a file an `include.path` names. A sandboxed command cannot create that state, since `config` is denied, so it matters where the caller already has it. Same on `main`.

**A bare-repository layout in a subdirectory, and sequencer state.** `HEAD`, `objects/`, `refs/` and `config` directly in a directory make it a repository git finds before the parent's `.git` when run from there; nothing looks for that layout. `<gitdir>/rebase-merge/git-rebase-todo` takes `exec` lines that `git rebase --continue` runs, and stays writable. Both as on `main`.

**The macOS profile is bounded by `ARG_MAX`.** It has no cap on entries of its own, but it travels on `sandbox-exec`'s command line (1 MiB): enough entries under a `.git/modules`, which a command can write, and every later command fails at spawn with `E2BIG`, untyped. The README said it had no cap; it now says this.

**Anchored at the working directory.** A repository under a second allowed write path outside the working directory gets none of the mandatory denies, as on `main`; the README's "within allowed write paths" overstated the scan's reach and is corrected. On Linux `git init` and `git clone` into a new subdirectory succeed, and the repository they make is not covered until the next wrap; the README listed both as broken.

**What decides a git directory outside a git-owned location.** A directory a pointer names that is not under a `.git/worktrees` or a `.git/modules` is still a git directory by one entry named `HEAD`, `objects` or `commondir`, so `x/.git` holding `gitdir: ../src` with an empty `src/HEAD` still aims the four denies, and a placeholder file, at `src`. Requiring two markers would stop recognising a bare repository's linked worktrees and a git directory used with `GIT_OBJECT_DIRECTORY`, and the same aim is available through a symlinked `.git/modules` entry with no write at all, so it is left.

**Sibling worktrees from inside a linked worktree.** Where the working directory is itself a linked worktree and the write roots also hold the main repository, the main `.git`'s `worktrees/*` and `modules` are not listed from the pointer end: only this worktree's own git directory and what its `commondir` names are. A command there can still rewrite another worktree's `commondir`. Listing from that end needs a visited set across repositories and a place in the degrade plan, and is a change of its own.

**A submodule name with one of git's own names in the middle.** Under a forged name segment, `vendor/lfs/foo` is still not reached, because `vendor/lfs` lists only `foo/` and is left at one listing like git's own `lfs`. Closing it means walking `objects` and `refs` fan-outs.

**A degrade that goes further than it must.** Where a step gives nothing back the next build is the last level. In one mixed layout (a repository inside the write roots sorting before one outside them, with the profile a few words over) stepping on would have stopped earlier; the profile still runs, with more submodules read-only than strictly needed.

## Merge order

`main` is merged in twice: at a7f016b, which put this on top of every change to the Linux deny loop and its producer that had landed by then (#542, #514, #503, #555, #536, #553, #535, #534, #562, #524 and #506), and again in this round, which takes #565 (CI on Bun 1.4.2, and the two tests it exposed - one of them `proxy-deny-violations`, which the branch's own base still failed), #564 and #561. No conflicts either time. A third merge in this round takes #570 and #571, which did conflict - five places where main and this branch had each added to the same one, each resolved by keeping both sides (the fullest case: main deleted `hasRipgrepSync` as dead while this branch added `RipgrepError` beside it, so the function goes and the class stays, main's deletion having taken the `whichSync` import with it). What that merge newly makes checkable is in the diff too: `GitMetadataError` and `SubmoduleWalkBudgetError` added to #570's pinned export list, and three mechanical type errors in this branch's own test files, which #571 put under `tsc` for the first time. `test/sandbox/seccomp-filter.test.ts` does not typecheck on main either, so main is red on every non-Windows leg as this is written; the merge repairs it with the one non-null assertion the same value already carries two lines above, and touches nothing else of that file. Nothing is owed ahead of this.

## Testing

Closest to real: `test/sandbox/mandatory-deny-paths.test.ts` drives each shape through the real wrapper, under bubblewrap 0.11 on Linux and Seatbelt on macOS, with a positive control wherever a failure alone would prove nothing. The parity claim is held by `test/sandbox/git-pointer-parity.test.ts`, a differential test against real git: each pointer is resolved twice, through this code and through `git rev-parse --absolute-git-dir` / `--git-common-dir`. Wherever git follows a pointer, that directory's `hooks`, `config`, `config.worktree` and `commondir` must be denied, or the wrap must have been refused; denying more passes, denying less fails. Also driven by hand through the `srt` CLI on Linux and `sandbox-exec` on macOS.

- The latest round, one new file each: `linux-work-tree-guard.test.ts` (an entry linked at the checkout, at a second write root, at a directory holding the checkout, and past the argument cap; the narrowed refusal; live two-command arms where the second command still writes the project), `mandatory-deny-modules-walk.test.ts` (a forged marker in a name segment, for each marker and each of git's own names; a 256-directory `objects` is not descended; live), `mandatory-deny-worktrees.test.ts` (real `git worktree add`; `HEAD` moved aside; a bare `main.git`; 900 worktree directories under the cap; live), `linux-git-dir-whole.test.ts` (mode 0555, both orders, live), `linux-git-redirect-placeholder.test.ts` (the name swapped for a symlink, a hard link or a FIFO after each touch in turn; the victim's bytes and mode unchanged), `mandatory-deny-time-budget.test.ts` (rebuild count; a chain of 800 detours across 48 git directories and 40 hard-linked large pointer files against a small budget). Each new test fails at 4750e50 except the ones written as guards, and each fix was reverted in turn to see its tests fail.
- Parse edges, unit-tested: both sides of the 1 MiB bound, NUL truncation, CRLF, every refusal path, `..` after a symlink, loops, FIFOs.
- Chains that reach nothing: the hops a two-link loop, a self-loop, a loop entered after two ordinary hops and a name too long to exist give back, each named once; a loop confined to the git directory, whose hops name no holder; the macOS entries, which carry every visited link and never the directory holding them; and a pointer value that loops, which denies the pointer and the links and never the checkout or a write root. On Linux, the deny plan for each: a loop a subdirectory holds gains one read-only bind and no refusal, a loop the working directory holds is refused with the new reason, and a looping pointer value leaves the working directory unbound. Under bubblewrap: every way to make a loop a subdirectory holds reach a directory - unlink, rename, retarget, `mkdir` over it, write a file over it - is refused while project writes elsewhere succeed; and two wraps in a row, the first planting a looping `.git` pointer and the second still writing the project while that file stays denied. Against real git: what it does with a looping hooks link (no hook, then the hook, once one link is replaced by a directory) and with a looping pointer value (opens nothing, and does not fall back to the lexical fold).
- A scan killed inside its budget after printing a `(os error 13)` line is refused rather than read as a complete listing, and `RipgrepError` says whether the run was killed and by which signal.
- The parity corpus crosses prefix, trailer and path spellings with target shapes, plus 150 seeded random mixes.
- A deny aimed at the project: unit cases for the working directory, an ancestor of it, a second write root, a directory holding a write root (none denied) and an ordinary directory (still denied whole); and live two-command arms for the two ways to plant one - a `.git` naming the checkout, and a `.git/modules` entry linked at `../..` - where the second command still writes the project. A working directory that cannot be listed refuses with `deny_scan_failed` naming it. A directory that is readable but not searchable, and one another user owns at the scan's depth limit, both keep wrapping; the second is pinned with a fake rg reporting the EACCES, which is decisive whatever the local ripgrep does.
- A `modules` that is a symlink: unit cases for a git directory's own and for a submodule's, with the chain past it; the reviewer's own sequence under bubblewrap (remove the link, put a directory there, write the hook) refused with nothing on the host; and the cost, live - a command plants `ln -s nowhere .git/modules` and the next command finds `git status` working, `git commit` read-only and the hook unplantable.
- The walk: a hundred submodules each nested inside the one above - two hundred directories of `<name>/modules` - are all found and every level's `hooks` and `config` denied; a symlinked entry pointing back at the tree root ends that branch and the rest of the listing still stands; two thousand eight hundred symlink loops one command can plant deny nothing at all, and a submodule beside them is still found; two links into a directory this process cannot search are denied at that directory, once, rather than at the `modules` holding them; a deadline already past throws; one directory of fifty thousand entries against a fifty-millisecond deadline is stopped inside it rather than at the end of it; a wrap given one millisecond of scan budget is refused with `deny_scan_failed`, while a walk made after the scan gets a budget of its own and is not. Plus a Linux arm that builds the argv and, where bubblewrap can create its namespaces, runs the wrapped command: inside a git directory twelve levels down a write to `objects/` succeeds while `hooks/`, `config` and the hooks of its own submodule fail, and on macOS the same directories are denied by literal path with no deny over either git directory whole.
- Degrading, as a function of a plan and a level: a level that degrades nothing gives back the list exactly as it was; a degraded git directory takes its deny paths from wherever they came from, including a checked-out submodule's copy of them, and leaves the pointer file's own deny alone; an entry whose real path is outside a `modules` denied whole keeps its own deny; a repository with no submodules has nothing to degrade and nothing denied at its absent `.git/modules`; degrading runs out, which is what `too_many_arguments` is left for; and the warning names the repositories and counts every level. Then the real thing: a repository with 2000 submodule git directories wraps rather than being refused, its mount words stay under bubblewrap's 9000, the first submodules in sorted order keep their `hooks` denies and the last are bound whole, two wraps produce byte-identical argument files, the violation monitor's own list is the undegraded one, and the macOS entries for that same tree give up nothing. A repository of checked-out submodules is driven at 300 and 700, where it wraps under the cap, and at 1600, where it is refused with `too_many_arguments`; multi-segment (`vendor<n>/<name>`) and nested names are driven at the same pressure.
- Entries that are symlinks, unit-tested: a git directory with a symlinked `hooks` is denied whole and the target denied on its own, with the target named as what a degrade has to keep; a link that stays inside the git directory adds nothing to keep; a dangling one is denied where it lands, a chain of dangling links at the end of the chain; a loop leaves only the whole-directory deny; an entry whose `lstat` refuses is taken for a link; the `.git/modules` walk denies the directory holding a symlinked entry and, for a dangling one, where it lands; and the macOS entries for the same tree carry the link and the target and NOT the git directory whole. Degrading: a collapsed git directory keeps what its symlinked entry leads to, under a `modules` denied whole as well, and a kept deny is a mount the step does not give back.
- Chains with a hop in the middle, unit-tested: a two-link chain carries the entry, the landing and the link between them, with the link's holder named as what a bind has to take; a symlinked directory component on the way is a hop like any other; a hop inside the git directory is named but needs no directory of its own; a chain of dangling links carries the hop as well as the landing; a chain running into a directory this process cannot search leaves the git directory denied whole and still holds the hop it could read; a loop has no hop to hold. The macOS entries for a chain carry the entry, the hop and the landing and deny neither the git directory nor the directory holding the hop. On Linux the argv shows the directory holding the hop bound read-only where it is a subdirectory, the landing bound on its own and what a symlinked directory component leads to NOT bound, and where the working directory holds the hop it shows no read-only bind over it after its own write bind, with the warning naming the entry and the hop.
- Live arms under bubblewrap for a chain. Where a subdirectory holds the hop: `ln -sfn` of the hop, renaming it aside, unlinking it and creating a name beside it are all refused, the entry cannot be unlinked, a write through to the landing is refused, an ordinary project write and `git status` still succeed, and the host still resolves the chain where it did. Where the working directory holds it, the retarget lands - pinned as a fact rather than left to be found - while the warning is emitted and both ends of the chain stay refused.
- The chain a POINTER's value walks, unit-tested at the same points: one symlinked component, two links with a `..` after them (both links held, in the order the walk met them), a component with no `..` in the value at all, a link that leads nowhere yet, a link the walk cannot read past (the git directory fails closed on the deepest ancestor it can reach, and the link it did read is still held), a `commondir` value, which names itself as the source rather than the pointer, a loop, which has no hop to hold, and a checkout reached through a symlink, whose own link is where the repository lives and not a hop. The macOS entries for such a chain carry the pointer file, the link and the hooks it leads to and deny nothing whole. On Linux the argv shows the directory holding the hop bound read-only where it is a subdirectory, and where the working directory holds it no read-only bind over it at all, with the warning naming the pointer file and the link.
- Live arms under bubblewrap for a pointer's chain, on a repository `git init` made and whose `.git` was then turned into a pointer through a link: where a subdirectory holds the hop, retargeting it, renaming it aside, unlinking it and creating a name beside it are all refused, the pointer file cannot be rewritten, the hooks it leads to cannot be written, and `git status` and an ordinary project write still succeed; where the working directory holds it, the retarget lands while both ends stay denied.
- The macOS spelling rule, unit-tested against the entry list: a pointer whose value walks through a symlinked component carries the hop, the git directory as the pointer spells it AND the same four files at the landing, the absent `commondir` and `config.worktree` among them; a submodule behind a symlinked `.git/modules` entry carries both spellings the same way; a chain of symlinked ENTRIES carries the entry, each hop and the landing and needs no second spelling, each of those paths being a landing already; and an ordinary repository names every path exactly once. The first two fail at the previous head. A third pins the hop exception: a pointer reaching its git directory through a link denies the hop and the hooks at the landing and denies neither the directory the hop leads to nor the checkout. A macOS-only arm, `test/sandbox/macos-git-chain-landing.test.ts`, runs the real profile under `sandbox-exec` on the pointer shape: a hook written through the link spelling, the same hook written at the landing and a `commondir` created at the landing are each refused, `rm` and `mv` of the link are refused, and an ordinary write in the checkout succeeds. All three of those writes land at the previous head.
- Two parity cases hold git's own rule for a `..` after a symlink, against real git: with both candidates git directories it opens the kernel's landing, and with only the lexical one there it refuses the checkout.
- The degrade warning: a plan whose `.git/modules` is already denied whole says nothing when those submodules are degraded.
- Live arms under bubblewrap for both halves. In a repository with 700 submodule git directories, two of the collapsed ones having a symlinked `hooks` - one to a directory inside the write root, one to a path that is not there yet - a write through the link is refused, the landing cannot be created, `rm` and `mv` of the link are refused, and an ordinary project write still succeeds. In a repository where nothing is degraded, with a symlinked `.git/hooks` and a symlinked `.git/modules/lib`, the write through the link, the `rm` and the `mv` of the hooks link and the `rm` of the modules link are all refused, while a project write and `git status` succeed. Before the fix, on this branch's head and on v0.0.77, each of those attempts landed; the argv dump over 72 configurations differs only in the ones that hold a symlinked git entry.
- Live arms under bubblewrap. On a real repository with 700 submodule git directories made by script: a hook write in a degraded submodule fails, so does a hook write in one that kept its precise denies, `mv` of the degraded git directory fails, a write to an ordinary project file succeeds, `git status` succeeds inside the sandbox, and the host's `git status` still succeeds after the command. On a repository with 3000 of them, where the whole `.git/modules` goes behind one bind, a submodule entry that is a symlink out of the tree still has its hooks refused, and so does one under `modules`. And with the working directory's own 703 submodules being degraded, `git submodule add` inside a nested repository that has no `.git/modules` of its own still works, and the directory it made is a directory rather than a mount point left on the host.
- The scan's failure paths are driven by a fake `rg` that writes rg's own stderr byte for byte, so they do not turn on the uid: a run whose codes are all `ENOENT` carries on and mounts nothing at the vanished path; one reporting `EACCES` with nothing unreadable on disk is refused; one with an error code no deny stands in for, and one that says nothing at all, are refused with the raw line in the message; one outlives a short `timeoutMs`; one cannot be spawned; and an aborted signal comes back as the `AbortError` it is. Where the uid allows a directory to be unreadable, the same fake drives the name shapes that defeat reading paths out of stderr - a directory named `src: x` holding an unreadable `locked`, one named `.: x`, a name ending in a space, a name holding a newline, and a stderr mixing `ENOENT` with `EACCES` - and each time the unreadable directory is denied and nothing beside it is. The real `rg` closes the loop under bubblewrap: one wrapped command takes a directory named `a: b` to mode 000, and the next is still refused the nested repository beside it and refused the `chmod` that would open it again.
- The files git reads back have argument-level cases for a `commondir` absent, present, left empty by bubblewrap's own shape, left empty with write bits by a host git mid-write, and already holding the placeholder; for a caller's own deny of the same file spelled relatively and with a trailing slash; for an empty `config.worktree` (legitimate, so left alone unless it has bubblewrap's shape); for what the cleanup removes and what it must not; for a git directory at mode 0555, which is denied whole instead of refusing; for a wrap that throws after the files were written, which leaves nothing behind; and for the refusal when no placeholder can be written at all. The killed-wrap case is driven by a child process that wraps and then `SIGKILL`s itself, so the leftover is the wrapper's own rather than a hand-made file.
- Live arms under bubblewrap, on a repository with a submodule and a linked worktree: while a wrapped command runs, the host's `git status`, `log`, `worktree list`, `rev-parse --git-common-dir`, a commit and `git status` in the submodule all succeed, and both `commondir` files hold `.`; after the command the files are gone and git still works. A leftover `.` is read by git and removed by the next wrap; the empty `commondir` an older release leaves makes git refuse every command and is repaired by the next wrap. With `$TMPDIR` inside an allowed write path, `chmod` and an in-place write of the placeholder source both fail inside the sandbox with "Read-only file system" and git still resolves its own git directory; twenty wraps mint one store directory.
- `test/utils/ripgrep.test.ts` covers the NUL split, what a failed run carries back, that `--no-config` is on every run, and that a kill inside the time the run was given is reported as the signal it was rather than as a timeout. A unit case covers the violation monitor's fallback: a repository whose `commondir` is past the size read refuses every wrap, but leaves the monitor the cwd's plain deny paths - and no deny under a `.git` that is not there.
- The macOS bracket cases sit with #542's own in `test/sandbox/macos-literal-deny-brackets.test.ts`, which is why that file appears in this diff: under a working directory named `a[b/c]d`, a submodule git directory, a directory the modules walk could not list, and a pointer's target and its `commondir`'s target are each denied by subpath; no emitted regex carries a bracket that came off the filesystem; and the patterns with no path to enumerate, a nested repository's submodule and the `.git` pointer filter, are anchored at the escaped working directory. Three macOS-only cases run the profile under real `sandbox-exec`: a hook planted in a bracketed submodule git directory is refused, so is a write in the directory the walk could not list - whose mode is given back before the write, so that what refuses it is the sandbox - and a write elsewhere in that git directory still succeeds.

</details>

<details><summary>How the pointer parse matches git byte for byte</summary>

A regular file only (opened `O_NONBLOCK` and type-checked on the fd, so a FIFO cannot block the host), at most the 1 MiB git accepts, the 8-byte prefix at offset 0, trailing `\n`/`\r` stripped, the path ending at the first NUL. `commondir` is parsed the same way minus the prefix. For the `..`-after-a-symlink case, the lexical answer is `gitMetadataTargets` and the kernel's is `physicalPath`.

</details>

Supersedes #515, which closed when its head repository was removed. Same branch, same history, plus the last review round.


antdres:
Reviewed at a300d6f. I ran the tests and live probes under bubblewrap 0.9, both as root and as an unprivileged user. I agree with the direction: on the merge base (316e9f3) a sandboxed `echo x > .git/commondir` still succeeds, and this PR closes it. But I reproduced holes in three of the new mechanisms, and one of them is a new way for a command to break every command after it.

### Must fix

**1. Two concurrent wraps drop the `commondir` deny, which leads to code execution on the host.** `activeSandboxCount` (linux-sandbox-utils.ts:1577, `cleanupBwrapMountPoints` :1620) is per process. The `.git/commondir` mount point written by `gitRedirectMountPoint` (:322) is shared on disk. When one `srt` process exits, it unlinks the mount point while another process's sandbox is still running, and that sandbox's bind is detached.

```sh
# repo with no commondir; settings allowWrite: [<repo>]
srt -s s.json -- sh -c 'sleep 2' &
srt -s s.json -- sh -c '
  echo x > .git/commondir || echo refused          # refused: Read-only file system
  sleep 5
  ls .git/commondir                                # No such file or directory
  echo "$PWD/../evil" > .git/commondir && echo OK  # OK
'
```

The host's `.git/commondir` is left pointing at a directory the command controls. With a `core.fsmonitor` line in that directory's `config`, the host's next `git status` ran the command. The Known limits section covers this race as "stops that one deny applying", but that deny is what stops git's hooks and config being redirected, and agents routinely run commands in parallel. One way out is to leave the `.` placeholder in place rather than unlink it, since the next wrap already handles a leftover. Another is to coordinate cleanup across processes, for example with a lock or refcount file next to the store.

**2. A directory that is readable but not searchable breaks every later wrap, and the error is untyped (non-root only).**

```sh
mkdir -p a/nested && git -C a/nested init -q
srt -s s.json -- chmod 400 a
srt -s s.json -- echo ran
# bwrap: Can't create file at <cwd>/a/nested: Is a directory
```

`readdir(a)` succeeds and reports `nested` as a directory, so `walkScanDirectories` (:584) recurses into it. `readdir(a/nested)` then fails with EACCES, and `a/nested` is pushed as the unreadable directory. The host can't stat that path either, so bubblewrap is asked to put a file mount at a path that is really a directory. Every later wrap in the tree fails the same way, and only a `chmod` run outside the sandbox clears it. On main the same tree runs. Modes 000, 100 and 300 are handled correctly; only r-without-x hits this. A fix is to deny `dir` rather than `child` when `dir` lacks search permission. There is no test with uid != 0 for this mode.

**3. A symlinked `<gitdir>/modules` is not held.** `gitDirDenies` (mandatory-deny-paths.ts:274) checks `hooks`, `config`, `commondir` and `config.worktree` for being links, but not `modules`. Setup: `.git/modules/foo` is a git directory, and `.git/modules/foo/modules -> ../../../../nm`, where `nm/bar` is a git directory. Writes through the link to `nm/bar/hooks` and to `foo/hooks` are both refused. But this succeeds and leaves the hook on the host:

```sh
rm .git/modules/foo/modules && mkdir -p .git/modules/foo/modules/bar/hooks && echo 'echo PWNED' > .git/modules/foo/modules/bar/hooks/post-checkout
```

That contradicts "both halves are denied". Checking `modules` the same way would put the git directory in the whole-directory deny, as with a symlinked `hooks`.

### Worth fixing

**4. macOS.** A submodule git directory reached through a symlinked `.git/modules/<name>` is denied only under the spelling through the link, because `toLiteralPathEntry` keeps that spelling when the real path is outside the boundary (macos-sandbox-utils.ts:163). Seatbelt matches the resolved path, so the generated profile never names `<real>/hooks`. I checked this from the profile text only, since `sandbox-exec` wasn't available.

**5. Degrade loop cost.** When a step saves no mounts (the submodule git directories are outside `allowWrite`, and the overflow comes from elsewhere, such as a large `denyWrite`), the loop around `collapseFurther` (linux-sandbox-utils.ts:2409) rebuilds the profile once per step before it refuses. With `allowWrite=[repo/src]`, 3010 `denyWrite` files and N bare submodules, reaching `too_many_arguments` took 0.5 s at N=0, 9 s at 500 and 25 s at 1000, and at 3000 it was still going after 150 s. Stopping when a pass doesn't shrink the word count would bound it.

### Checked and fine

- The ripgrep fail-closed logic: the errno classification, NUL splitting, `--no-ignore` and `--no-config`.
- Hiding a repository by emptying `.git`, with `allowGitConfig` both on and off.
- A FIFO at a git metadata path.
- A linked worktree's real `commondir` is kept and bound read-only.
- The placeholder store stays read-only inside the sandbox even with `$TMPDIR` writable.
- The degrade never drops a deny it should keep: I checked 1000 submodules and 3200 submodules with symlinked entries live.
- An ordinary repo can still `git add` and `git commit` inside the sandbox.

The three changed test files pass (184 pass, 15 skipped as root). The full suite has the same failures as the merge base in this environment, and none are new.

---

PR #570, opened 2026-09-18
Make nine tests able to fail, and cover globToRegex and the public export list

ronleizrowice-ant:
## Summary

Nine tests cannot fail: five only print what they saw, three skip their assertions when a dependency is missing, one sweeps a corpus that never reaches the branch its comment claims. Two chokepoints have no coverage: the glob compiler all three backends' deny rules go through, and the package's export list. Tests only; no source changes.

## What was wrong

- `allow-read.test.ts`: the five cases under "adjacent aliasing behaviors" call a helper that `console.log`s a line and hold no `expect` between them. Two are read-deny bypasses: were a build to let a hard link reach a denied file, the logged word would change and CI stay green.
- `seccomp-filter.test.ts`: the first test wraps its assertions in `if (errors.length === 0)`, two return mid-body when the apply-seccomp binary does not resolve, and the five integration tests open with a guard that returns when a dependency is missing. With bwrap missing, all pass having wrapped nothing.
- `winsrt-paths.property.test.ts`: the corpus is five generated file names and the generator only selects members of it, so no sample holds `*` or `?`. The glob branch is never entered; the property compares the literal branch with itself.
- `aws-sigv4.test.ts`: every vendored request lists its headers lowercase and sorted, as does the one header the test adds, so the signer's lowercase and sort steps are never exercised. Deleting both leaves all 22 vectors green.
- `globToRegex`, the one place a glob deny becomes a pattern, had four cases: `*`, `**/`, `?` and a trailing `**`. The documented bracket form had none.
- Nothing checked the 49 values `src/index.ts` exports, and building declarations cannot notice one going missing.
- The mandatory-deny fixtures claim "ALL dangerous files" above nine names written out by hand, and import neither constant the backends read.

## Change

Each aliasing case asserts the direction this build shows; the two that depend on the machine are decided at collection and skip where they cannot run. The dependency check becomes a `beforeAll` that asserts, and the swept corpus gains two patterns. Each SigV4 vector signs again with its header set reversed and upper-cased, and the fixture count is pinned exactly. `globToRegex` gains cases for a digit range, a letter range, both negation spellings, an unclosed `[` and a stray `]`, plus a property checking it against a hand-written backtracking matcher over generated inputs. A new test compares the entry point's sorted export names with a checked-in list; the deny fixtures and their blocks loop over the constants.

## Who sees a difference

A machine missing bwrap, socat or the built apply-seccomp binary now fails the seccomp suite instead of passing it. CI installs and builds all three. Nothing changes for users of the package.

## Not done here

- A path component spelled like one of the compiler's internal placeholders comes back as a wildcard. Marked failing rather than fixed, so it reds the suite once it is.
- A bracket set is never negated: both spellings land in the set as members. Pinned, not changed.
- Type exports are erased before the export test sees them; catching one needs a declaration snapshot the package does not build.

## Testing

Each new or changed assertion was shown to fail by mutating the source and reverting: escaping brackets in the compiler, turning `**/` into `.*`, dropping the signer's sort, reporting a missing dependency, normalising differently in the glob branch, dropping an export, skipping a deny entry in a backend. Two controls confirm the gaps: with the sort dropped the old vectors all pass, and with a dependency missing so does the old integration block. `tsc --noEmit`, `eslint` and `prettier --check` clean; full `npm test` matches a clean checkout of `main`.


antdres (approved):
I merged this with current main locally (clean) and ran it on Bun 1.4.2 with bwrap, socat and apply-seccomp installed. All touched files pass. Then I broke the source each test guards. Every repaired test now goes red where main's version stayed green: SigV4 without the sort or lowercase step, a missing bwrap, the Windows glob branch normalising differently, a removed export, and a new `DANGEROUS_FILES` entry that Linux skips. Fixing the `__GLOBSTAR__` placeholder turns the `it.failing` case red, as intended.

One non-blocking suggestion: turning `**/` into `.*` in `globToRegex` is caught only by the random fast-check property, which failed in 13 of 30 runs. A regression there would look flaky. A fixed case would make it deterministic, for example `/tmp/**/b` must not match `/tmp/ab`. The property stayed green over 50,000 runs on unmodified code, so it doesn't produce false reds.

---

PR #571, opened 2026-09-18
Correct five false statements, and remove dead and duplicated code

ronleizrowice-ant:
## Summary

Statements the repository makes about itself that are not true, plus code that
cannot run and code that says one thing twice. One behaviour change: `srt
--version` reported 1.0.0.

## What was wrong

| Where | What it said | What is true |
| --- | --- | --- |
| `src/cli.ts` | `--version` reads `npm_package_version` or `1.0.0` | npm exports that only to its own lifecycle scripts, so an installed `srt` always printed 1.0.0 against a package at 0.0.77 |
| `README.md` | Linux cannot block a mandatory deny path that does not exist yet; Linux violation monitoring is future work | absent paths get a read-only `/dev/null`, or an empty directory at the first missing component, removed afterwards; the monitor ships and `initialize()` starts it |
| `CredentialsConfigSchema` | `mask` "will be added in future releases" | it shipped; the same file enumerates and validates it |
| `windows-sandbox-utils.ts` | 30s is "the same budget as windowsTrustCa" | that call has always passed 60s |
| `winsrt.test.ts` | defers the read half to a SYNCHRONIZE-strip fix, "until then" | the strip is in `acl.rs`, and the smoke script asserts that half today |
| `package.json` | `typecheck` runs `tsc --noEmit` | that is the src-only project; `tsconfig.test.json` exists to cover `test/` and nothing ran it, CI ran no typecheck at all, and bun strips types |
| `integration-tests.yml` | a step "verifies bwrap can create namespaces"; the Windows legs run the TS suite | both arms echo, so it cannot fail, and it probes a different surface from the gate suites skip on; the legs ran two files, so the only property suite over the Windows path pipeline ran nowhere |

## Change

Each of those is corrected where it sits. Alongside them:

- Wiring the test project into `typecheck` surfaced 70 errors; all are fixed.
  Most were one shape: a header object a server callback records, whose write
  the compiler cannot see, so the reset a test does before its request narrowed
  the variable to `undefined`. Those go through a small `captured` slot now.
- Dead code: two branches behind an equality check that returns first, three
  `process.arch` cases the closed union lacks, a `statSync` result tested for
  `isSymbolicLink`, a jar lookup with no caller but its own test, an
  unreferenced `hasRipgrepSync`.
- Said twice: the write-glob strip, pasted three times under two log strings;
  `RipgrepConfig`, re-spelled by hand seven times, each dropping `argv0`; the
  install timeout, in prose three times over an inline literal; a stub-skip
  comment longer than its function; a seccomp walkthrough that had already
  diverged from the README's.
- `dtolnay/rust-toolchain` and `Swatinem/rust-cache` are pinned by commit in
  both workflows; the container arm pins bun and takes node from the image
  rather than a second piped installer.

## Who sees a difference

`srt --version` reports the installed version. The rest is comments, types,
dead code and CI: no behaviour change, no renamed export.

## Not done here

- The Rust compiler version for the elevated Windows helper still floats; only
  the action code is pinned.
- `test/utils/` stays off the Windows legs: those suites assert a POSIX tool
  layout, so adding them means gating nearly every case.
- `whichSync` still spawns per wrapped command; memoizing it is a behaviour
  change and the suites mock the module. The vendored-artifact lookup is still
  written twice, in the seccomp and jar modules; sharing it means a shared
  cache and a new home for it.
- Deliberately untouched: the package `exports` map, strict config schemas,
  validating the config in `initialize()`, Windows mandatory denies,
  documenting `--control-fd`.

## Testing

`tsc` on both projects, `eslint` and `prettier --check` clean on every changed
file. Full `npm test` on Linux against this branch and against a clean
checkout of `main` on the same machine: the same tests fail on both, test for
test. `srt --version` prints `0.0.77`. Running config-validation on the
Windows legs turned up one assertion that is POSIX-only, since a bracket
spelling is a literal path there; it is gated to the platforms it holds on.


antdres (approved):
Checked each removal against the tree: the `/private/tmp`/`/private/var` branches and `resolvedIsSame` sit behind the equality early return, the three `process.arch` cases are outside Node's union, `statSync` never reports a symlink, and `hasRipgrepSync` / sync `getJavaProxyAgentJarPath` have no callers and aren't exported from `src/index.ts`. The `normalizePathForSandbox` rewrite is equivalent.

Verified locally: both typecheck projects clean (main has 70 errors under `tsconfig.test.json`), lint clean, and `node dist/cli.js --version` prints 0.0.77 with `npm_package_version` unset. The pinned commits match `rust-toolchain`'s `stable` head and `rust-cache` v2.9.2. The changed suites show no failures beyond ones main already has in the same environment.

Non-blocking: with no `exports` map in package.json, the two removed functions were deep-importable from `dist/`, so an out-of-tree caller could notice. Fine at 0.0.x. This also supersedes #217 and #135.

---

PR #573, opened 2026-09-19
globToRegex: negate bracket sets, and compile from tokens so no path text can become a wildcard

ronleizrowice-ant:
## Summary

One compiler turns a glob into a regular expression for all three backends: the `regex` filters of a macOS profile, the Linux walk that expands a read deny, and the Windows expansion. It never negated a bracket set, though the syntax it documents is gitignore's, so a rule written `[!x]` or `[^x]` covered the opposite names. And it parked `**` under the strings `__GLOBSTAR__` and `__GLOBSTAR_SLASH__` while it rewrote `*` and `?`, restoring them by name, so a path component spelled that way came back as a wildcard.

## What was wrong

| A rule written | compiled to | which matches |
| --- | --- | --- |
| `/data/[!a]*` | `^/data/[!a][^/]*$` | names starting with `!` or `a` |
| `/data/[^0-9]x` | `^/data/[\^0-9]x$` | `^x` and `0x` to `9x`: the opposite set |
| `/tmp/__GLOBSTAR__/x` | `^/tmp/.*/x$` | every directory, not the one named |
| `/tmp/__GLOBSTAR_SLASH__x` | `^/tmp/(.*/)?x$` | anything but the file it names |
| `/tmp/a[b[c` | `^/tmp/a\[b[c$` | nothing: not a regular expression |

## Change

The pattern is read once, left to right, and the regex is emitted as it goes, so nothing is parked under a marker and no text in a pattern can be taken for one. A `!` or `^` first inside a bracket set negates it, and a negated set never matches a separator, being one character of one name. `/data/[!a]*` compiles to `^/data/[^/a][^/]*$`.

Two corners of the bracket syntax come with that. A `[` that opens no set is escaped wherever it stands rather than only the first one, and a set with no members (`[]`, `[!]`) is not a set, so its brackets are characters of the path. `]` is still never a member: no spelling of it inside a set reads alike to a JavaScript regular expression and to the sandbox profile's, which the same string also becomes.

A `-` that is a member of a set (first or last in it, straight after a range, or an end of one) is written first among the members and never behind a backslash: `[a-]` as `[-a]`, `[!-a]` as `[^-/a]`, `[!--0]` as `[^-/.-0]`. That is the one spelling both engines read alike. The profile's engine takes a backslash inside a set for a member, so the earlier `[^/\-a]` was "not `/`, and not `\` to `a`" there: narrower than written, and it let `-` through. Writing the `-` last is no way out, because that engine rejects a set ending in one character and a `-` (`[a-]`, `[^/a-]`) as an unterminated bracket expression, and the whole profile with it.

The walk that follows a read-deny glob a name at a time fenced off two shapes it could not read back from the rewriting: one spelling a placeholder, one with a second unclosed `[`. Neither is a shape of its own now, so both fences are gone. The README names the negated set and what a deny written that way covers on macOS, and drops the fence that went with them.

## Who sees a difference

- A deny written `[!x]` or `[^x]` denies what it says; it denied the opposite set of names, plus the literal `!` or `^`. An allow written that way changes the same way.
- A path component spelled `__GLOBSTAR__` or `__GLOBSTAR_SLASH__` is that name and nothing else.
- On macOS only, measured with `sandbox-exec`: the profile's own regex engine reads a negated set whose members sit next to `/` in the character order (`[!0-9]`, `[!.]`) as any character but a separator, wherever the rule carries the optional tail a deny always gets. Such a deny covers the names it should and the ones it lists as well: wider than written. One thing goes the other way, for one odd name, and is older than this change: the backslash this compiler writes before `.`, `+` and the like inside a set is one more member to that engine, so a negated set holding one also excludes a backslash, and a deny `[!.]x` does not cover a file named `\x`. A set that holds a `-` of its own is now written without those backslashes; the others are left as they were, so that they still compile to the string `main` gives.
- On macOS a rule with a set that ends in one character and a `-` (`[a-]`, `[ab-]`) now loads. `sandbox-exec` rejected the whole profile for it before, on `main` and in the released versions, so every command failed.

## Not done here

Nothing outside the compiler, those two fences and the README's syntax. A bracket set holding a wildcard (`[*]`) is still rewritten like any other, and still matched against whole paths; one that also starts with a `-` keeps the old `\-`, since what it compiles to is not a set at all and every cheaper spelling turns some pattern that compiles today into one that does not. A range whose ends are the wrong way round (`[z-a]`) still gives a regex that does not compile, as before.

## Testing

`tsc --noEmit`, `eslint` and `prettier --check` clean on the changed files; every suite that consumes the compiler green, and the full suite matches a clean checkout of `main` on the same machine.

The differential property from #570 now generates negated sets and both placeholder spellings. Swept exhaustively, 15,559,840 pattern and path pairs agree with the reference matcher where the old compiler diverges on 305,509; of 16,109 patterns compiled by both, 15,097 came out byte for byte the same before the `-` rule, the rest in the classes above. A set holding a `-` of its own is now a fourth class that differs from `main` (`[a-]` is `[-a]`). For that rule on its own: every set body of up to five characters from a fifteen-character alphabet, in all three spellings, is 2,440,845 patterns; 2,247,790 compile to the string they did before it, and the 193,055 that differ are every one a wildcard-free set holding its own `-`. No JavaScript match differs between the two on 95 characters and four path shapes, and the reference matcher agrees with both on 14,528,388 pattern and path pairs. A rule checked over every generated set holds for all of them where the earlier output broke it 166,180 times: a member `-` only first, none before the closing `]`, no backslash but a backslash's own.

Both backends were driven live, each against `main` as a control. Linux, through the walk and bubblewrap: a deny `keys/[!a]*` leaves `a.key` readable and refuses `b.key` and `c.key`, where `main` refuses `a.key` alone; a deny of a directory really named `__GLOBSTAR__` covers it and leaves its sibling readable. macOS, `sandbox-exec` over the emitted profile: that deny refuses the names outside the set and not those in it, where the old spelling refuses the inverse, and `__GLOBSTAR__/**` covers that directory alone where the old spelling covered every sibling. For the `-` rule, `[^-/a]`, `[^-/.]`, `[^-/a-c]` and `[^-/.-0]` were each run under `sandbox-exec` and exclude exactly the characters JavaScript says they do, `[^/a-]` and `[a-]` were refused there, and `[-a]` and `[a-c-]` accepted. The positive forms the rule now writes for a range (`[-.-0]`, `[-+-,]`) and the lone `-` have not been through `sandbox-exec` yet; the tests pin their exact strings so that they can be.

## Merge order

#570 is on main, and `main` is merged in here. Nothing is waited on.


README.md
```diff
@@ -422,6 +422,9 @@ Paths support git-style glob patterns on macOS, similar to `.gitignore` syntax:
 - `**` - Matches any characters including `/` (e.g., `src/**/*.ts` matches all `.ts` files in `src/`)
 - `?` - Matches any single character except `/` (e.g., `file?.txt` matches `file1.txt`)
 - `[abc]` - Matches any character in the set (e.g., `file[0-9].txt` matches `file3.txt`)
+- `[!abc]` and `[^abc]` - Matches any character outside the set, and never `/` (e.g., `file[!0-9].txt` matches `filea.txt` but not `file3.txt`)
```
antdres, line 425: This one probably needs a release note. Any existing rule written `[!x]` or `[^x]` now covers the opposite set, allow rules included. An `allowRead` or `allowWrite` like `/data/[!a]*`, which used to match only names starting with `!` or `a`, now matches almost every name in that directory. Someone upgrading would get a wider allow without changing their config.

For context, I compared the old and new compiler on 529 patterns: 44 realistic denies plus every glob string in the repo's tests and docs. Only the negated-set, empty-set and second-unclosed-`[` cases changed, so this is the one user-visible change worth calling out.

---

src/sandbox/sandbox-utils.ts
```diff
@@ -1114,27 +1118,94 @@ export function encodedCommandFromProxyUser(
...
+        continue
+      }
+      regex += set.negated ? '[^/' : '['
+      inSet = true
+      i = set.members
+      // A `-` first among the members would read as a range with that `/`.
+      if (set.negated && globPattern[i] === '-') {
+        regex += '\\-'
+        i++
+      }
```
antdres: Has this `\-` escape been checked under `sandbox-exec`? The live macOS run in the description covers `[!a]`, which has no leading `-`. In JavaScript, `[^/\-a]` reads as "not `/`, `-` or `a`". If the profile's regex engine treats a backslash inside brackets as a literal character, the way POSIX bracket expressions do, it would read `\-a` as the range from `\` to `a`. A deny like `[!-a]x` would then fail to cover `_x` and would match `-x`, which is narrower than written.

It's a narrow corner, but a deny with `[!-a]` or `[!-.]` under `sandbox-exec` would settle it. If the escape doesn't hold there, putting the `-` last among the members would avoid needing it.

---

antdres (commented):
Two non-blocking notes inline. The rewrite itself looks right to me. I fuzzed it against a separately written reference matcher, and the Linux walk agrees with the full-path regex on patterns that spell `__GLOBSTAR__` or carry a second unclosed `[`, so removing the two checks there is safe.

---

antdres (dismissed):
Approving. The two inline notes (a release note for the `[!x]` flip, and a `sandbox-exec` check of the `\-` escape) don't block this.

---

PR #574, opened 2026-09-19
linux: turn four comment-only invariants into code

ronleizrowice-ant:
Stacked on #569 (`fix/linux-nested-repo-hooks-depth`) and carrying its commits until that merges. Seven commits are this branch's: four invariants, a comment cleanup, and a fifth invariant with the commit that takes it back out.

## Summary

Four rules this code states in comments and enforces nowhere become code. Nothing the sandbox does changes: the emitted argv is what #569's head emits, and the only new refusal is a pair of nested sentinels at wrap time, which nothing reaching that today can produce.

## What was wrong

- No registered sentinel may be a substring of another, says the registry, and the body-substitution scan is sound because of it. Nothing checked.
- Three comment blocks say no path is resolved before the mandatory-deny scan: a realpath taken ahead of it would miss a symlink retargeted meanwhile. True only because every resolution hangs off a lazy memo called later, and a memo reads as caching.
- The deny pre-pass and the deny loop are asked, in prose in one of them, to keep in lockstep: six predicates written out twice by hand. A directory the pre-pass records but the loop does not re-bind read-only suppresses a stub the sandbox needs.
- The buffered deny binds were a flat `string[]` five sites pushed three words into, with one consumer storing an index into it and two maps keyed by destination kept in step by hand.

## Change

One commit each. The sentinel check throws naming both credentials. `canonicalForm` throws when it runs before the scan: the generator sets a latch once the scan is behind it, or at once where there is no write config to scan for, and hands it to every pass it builds. `classifyDenyPath()` answers what an entry is - a skip, an unresolvable symlink, a symlink to mask, or an absent path, a directory or a file - and both passes switch over it exhaustively. Each deny bind is an entry carrying its source, destination, raw spelling and git redirect, and the emission is the only place that knows the words. Also here: a dead `try`, two leftover `// fs already imported` notes, and three comments labelled `Fix 1` / `Fix 2` that number nothing.

## Taken back out

A fifth commit gave each wrap a token, so that one cleanup could not release another's deferral, and added `wrapWithSandboxScoped()` to pair the two. Review found the tokens behave as the counter did: a cleanup called without one still frees the oldest wrap, and nothing held in one process's memory can see a second `srt` process adopt a live mount point and unlink it. The kernel can, which is #584: each wrap names its mount points in a manifest bubblewrap holds an fcntl lock on for the sandbox's lifetime, and cleanup becomes an idempotent garbage collect with no counter and no pairing to keep. That commit is reverted here, so this branch adds no public API for a contract #584 removes, and leaves the cleanup path as #569 has it.

## Not done here

Extracting the path resolver out of the filesystem-args generator; that generator's split is #569's. `getLinuxDependencyStatus`, whose deletion would take its own tests with it. The two socat bridges `initializeLinuxNetworkBridge` builds twice.

## Testing

`tsc --noEmit` on both projects, `eslint` and `prettier --check` clean. New tests: a nested sentinel is refused and leaves the registry untouched; the deny mounts are the same set whichever order the deny paths are listed in; an upgraded placeholder leaves the binds buffered around it alone.

The argv for 84 configurations - denies by glob, symlink, absence and placeholder, masked credentials, worktree pointers, symlinked git entries, chain and pointer hops, and repositories of up to 700 submodules, seven of them past the command line and into bubblewrap's argument file - matches #569's head: 53042 words, 82 cases byte for byte, and the two whose scan order varies run to run as the same set of words. A differential fuzz over 18,000 random trees agrees, with no more order noise than #569's head shows against itself. The full `bun test` matches that head's run on the same box.

## Merge order

After #569, whose commits this branch carries until that merges. Independent of #584, which changes the cleanup path this branch now leaves alone.


antdres (changes requested):
Requesting changes on one commit, d0c1ba0 ("a token per wrap"). The other five (sentinel nesting, the scan latch, `classifyDenyPath`, the deny binds as entries, and the dead-code cleanup) look good to me. I checked the "argv unchanged" claim independently: a differential fuzz over 5,600 random trees (symlinks, dangling links, leftover 0444 mount points, `.git`/`commondir`/`hooks` names) gave byte-identical argv before and after `8442731` and `d850089`. The full `bun test` run has the same failure set as #569 + main on the same box.

## What the cleanup is protecting

A deny on a path that doesn't exist yet needs a mount point file on the host. Unlinking that file while a sandbox is still bound over it detaches the mount inside that sandbox. The denied path can then be created, and whatever gets written there lands on the host. So the rule is: **never unlink a placeholder while any live sandbox relies on it.** Removing ghost files is secondary.

## Why d0c1ba0 doesn't enforce that

It infers whether a sandbox is still running from what callers report, held in one process's memory:

1. **The tokens act exactly like the old counter.** `release()` calls `cleanupAfterCommand()` without its token, so `releaseWrap(undefined)` frees the *oldest* wrap. The only tokened caller is the wrap's own `catch`, which frees the same count as before. With the `if (wrap !== undefined)` branch deleted, the new "cleans up once per wrap" test still passes. With the `released` flag deleted, it fails. So the fix is the flag on the handle. An embedder that calls `cleanupAfterCommand()` twice for a legacy `wrapWithSandbox()` wrap still deletes the mount points of a live `wrapWithSandboxScoped()` wrap. I confirmed that with a probe on this head.
2. **Nothing covers a second srt process.** Process B finds A's live placeholder, treats it as a leftover (`isStaleBwrapMountPoint`), adds it to its own `bwrapMountPoints`, and unlinks it when B exits. No in-memory scheme can see this. I reproduced it on #569 with two concurrent `srt` runs in one repo: the survivor lost its `.git/commondir` deny. #569 adds more shared placeholders inside `.git`, so this exposure grows with it.
3. **A caller that crashes, calls cleanup early, or never calls it** breaks the scheme the same way.

## Suggested fix: have the kernel report whether the sandbox is running

bubblewrap already has `--lock-file DEST`, which holds a lock for exactly the sandbox's lifetime and is released on any exit, SIGKILL included. With bwrap 0.9, running `bwrap --ro-bind / / --dev /dev --lock-file $L sleep 3`, the host's non-blocking exclusive `lockf` on `$L` is refused while the sandbox runs and granted after it exits.

Sketch:
- Each wrap writes a small manifest of the placeholders it relies on, in a per-user runtime directory (0700). It takes a shared lock on the manifest at wrap time, before bwrap exists, and holds it until `release()` or process exit. It also adds `--lock-file <manifest>` to the bwrap invocation, so the lock is covered from wrap time to sandbox exit.
- Cleanup, from any process and any number of times, unlinks a placeholder only if no *locked* manifest lists it (unlocked manifests are stale and get collected). Do the check and the unlink while holding the manifest directory's lock, so a wrap that starts concurrently can't slip in between them.
- With that in place, a double, missing or early cleanup is harmless, and so are crashes and concurrent srt processes. `cleanupAfterCommand()` becomes an idempotent garbage collect, and the counter, the tokens and the "exactly once per wrap" contract can all go.

It costs one `--lock-file` per wrap and no extra mounts, so it doesn't eat into bwrap's argument cap. Placeholders also stop lingering while unrelated wraps are outstanding.

## Asks for this PR

- Drop the token set from d0c1ba0, or wire the token through `release()`. As written, the commit message and README describe a guarantee the code doesn't provide.
- Keep `wrapWithSandboxScoped()` if you like it as a convenience, but document that it only guards against a doubled `release()` on its own handle.
- Do the lock-based fix as its own PR, ideally before #569 lands.

---

Issue #210, opened 2026-04-06
Add allowMachRegister config option (needed for Chromium/Playwright)


dontgitit:
## Problem

`allowMachLookup` was added in #83 to allow sandboxed processes to look up Mach services. However, Chromium-based browsers (Playwright, Puppeteer) also need to **register** Mach ports via `bootstrap_check_in`, which requires the separate Seatbelt operation `mach-register`. There is currently no way to allow this.

When launching Chromium through the sandbox (e.g., Playwright MCP server wrapped in `srt`), it crashes with:

```
[FATAL:base/apple/mach_port_rendezvous_mac.cc:159] Check failed: kr == KERN_SUCCESS.
bootstrap_check_in com.google.chrome.for.testing.MachPortRendezvousServer.596: Permission denied (1100)
```

Even with `allowMachLookup: ["com.google.chrome.for.testing.*", "org.chromium.crashpad.*"]` configured, the browser cannot start because `mach-lookup` and `mach-register` are distinct Seatbelt operations.

## Background

- `mach-lookup`: allows a process to **find** an existing Mach service (client side)
- `mach-register`: allows a process to **register** a new Mach service via `bootstrap_check_in` (server side)

Chromium's multi-process architecture requires the browser process to register `MachPortRendezvousServer` so renderer/GPU child processes can connect to it. This is core IPC — there's no flag to disable it.

`(allow mach-register ...)` is a supported Seatbelt operation, used in many Apple system sandbox profiles (e.g., `application.sb`, `system.sb`, `com.apple.homeenergyd.sb`).

## Proposed solution

Add an `allowMachRegister` config option parallel to `allowMachLookup`:

```json
{
  "network": {
    "allowMachLookup": ["com.google.chrome.for.testing.*", "org.chromium.crashpad.*"],
    "allowMachRegister": ["com.google.chrome.for.testing.*", "org.chromium.crashpad.*"]
  }
}
```

This would generate Seatbelt rules like:

```scheme
(allow mach-register (global-name-prefix "com.google.chrome.for.testing."))
(allow mach-register (global-name-prefix "org.chromium.crashpad."))
```

The implementation should mirror `allowMachLookup` — same wildcard suffix convention, same place in `generateSandboxProfile`.

## Use case

Running Playwright MCP server through `srt` for browser automation in Claude Code. Currently the only workaround is removing the `srt` wrapper entirely, which drops all sandbox protections.

## Related

[...]

PR #598, opened 2026-09-22
feat(macos): allowMachRegister, so Chromium can check in its Mach ports

v0ropaev:
Fixes #210.

## The gap

`mach-lookup` and `mach-register` are distinct Seatbelt operations, and the profile only ever emits the first. Both the hardcoded Apple-service block and the user-facing knob at `src/sandbox/macos-sandbox-utils.ts:1022-1030` emit `mach-lookup`, so a sandboxed process can *find* a Mach service but never *publish* one — `bootstrap_check_in` falls through to `(deny default (with message ...))` at `src/sandbox/macos-sandbox-utils.ts:970` however `allowMachLookup` (`src/sandbox/sandbox-config.ts:769`) is spelled.

Chromium's multi-process design needs the browser process to register `MachPortRendezvousServer` before renderer, GPU and Crashpad children can reach it, so Playwright and Puppeteer die at startup under `srt` and the only workaround is dropping the sandbox wrapper entirely.

Reproduced through this repo's own generator on `main` (ddbeb74), macOS 27 arm64, with a small C probe that calls `bootstrap_check_in`, wrapped by `wrapCommandWithSandboxMacOS` with `allowMachLookup: ['com.example.f08159d6.*']`:

```
bootstrap_check_in(com.example.f08159d6.probe) -> kr=1100 (Permission denied)
```

`1100` is exactly the code in the issue's crash log. Unsandboxed the same probe returns `kr=0`.

## The change

`network.allowMachRegister`, the server-side counterpart of `allowMachLookup` — same trailing-wildcard spelling, same place in `generateSandboxProfile`:

```jsonc
{
  "network": {
    "allowMachLookup":   ["com.google.chrome.for.testing.*", "org.chromium.crashpad.*"],
    "allowMachRegister": ["com.google.chrome.for.testing.*", "org.chromium.crashpad.*"]
  }
}
```

```scheme
; User-specified XPC/Mach services the sandbox may register
(allow mach-register (global-name-prefix "com.google.chrome.for.testing."))
(allow mach-register (global-name "org.chromium.crashpad.ReportHandler"))
```

The two options share their moving parts instead of duplicating them:

- **`machServiceNameSchema`** (sandbox-config.ts) — the trailing-`*` refinement and its message, previously inline in `allowMachLookup`, now defined once and used by both.
- **`machServiceRules(operation, names)`** (macos-sandbox-utils.ts) — the name-to-rule mapping, previously inline in the `allowMachLookup` block.

One wildcard convention is precisely what the emitter assumes, so two copies could drift while a single schema kept accepting both spellings. This is the only existing code the PR modifies: 5 lines in the emitter, 12 in the schema.

Measured against the patched generator:

| config | result |
| --- | --- |
| `allowMachLookup` only | `kr=1100 (Permission denied)` |
| `allowMachRegister: ['com.example.f08159d6.*']` | `kr=0 (SUCCESS)` |
| `allowMachRegister: ['com.example.f08159d6.probe']` (exact) | `kr=0 (SUCCESS)` |
| `allowMachRegister` for an unrelated prefix | `kr=1100 (Permission denied)` |

The last row matters: this is an allowlist, not a blanket opening of the operation.

## Security note

Registration is the wider grant of the pair, so it gets a `### Security Limitations` entry (where `allowUnixSockets` and `allowAppleEvents` already have one) and `allowMachLookup` does not. Looking up only reaches outward from the sandbox; registering publishes a name into the login session's bootstrap namespace, where processes **outside** the sandbox resolve it. Measured, not assumed — a sandboxed holder registered a name and an unsandboxed lookup found it:

```
[sandboxed]  register com.example.f08159d6.squat -> kr=0
[outside]    lookup   com.example.f08159d6.squat -> kr=0 (FOUND) port=0x1d03
[outside, after the holder died]                 -> kr=1102 (Unknown service name)
```

So a broad pattern lets a sandboxed command claim a name whose real server has not started yet and receive the IPC its clients intended for it. The bound is in the entry too, and is also measured: a name already held by a running service **cannot** be taken over (that attempt returns `1100`). The entry says to scope patterns to what the command genuinely registers rather than `"*"`.

## Verification

macOS 27 arm64, node v24.8.0, bun 1.3.14. (`bun install --frozen-lockfile` fails here because bun migrates `package-lock.json`; plain `bun install` is fine.)

- `bun run typecheck` -> rc 0
- `bun run lint:check` -> rc 0
- `npx prettier --check` on all six touched files -> clean
- `bun test test/config-validation.test.ts test/sandbox/macos-seatbelt.test.ts` -> **156 pass / 0 fail**. Baseline on those files: **150 pass** (+6 new tests).
- `bun test` -> **1143 pass / 689 skip / 61 fail**. The 61 are pre-existing: stashing the patch and re-running on ddbeb74 gives **1137 pass / 689 skip / 61 fail**, and diffing the sorted failing-test-name lists prints nothing — identical sets. They sit in `--control-fd` and the TLS-terminating-proxy / credential-masking suites, none of which this touches.

Tests added, mirroring the existing `allowMachLookup` pairs: `global-name` vs `global-name-prefix` emission; a guard that a register-only name gets **no** `mach-lookup` rule; the profile compiles and runs with `['com.example.service', 'com.example.prefix.*', '*']`; and accept/reject config-validation cases.

Mutation-checked, and 4 of 5 mutations are caught. Neutering the shared refine fails **4** tests — both lookup *and* both register rejects, which is the evidence that the factoring is genuinely shared; giving `allowMachRegister` a bare `z.string()` fails 2; pointing the emitter at `mach-lookup` fails 2; deleting the profile block fails 1. The survivor is deleting the `sandbox-manager.ts` pass-through — nothing in the suite covers the config → manager → profile path for *any* macOS option (`allowMachLookup`, `allowAppleEvents`, `allowLocalBinding` and `allowUnixSockets` are all tested by calling `wrapCommandWithSandboxMacOS` directly), so I matched the existing level rather than introduce a new test harness here.

## Left out, on purpose

- **No README config-reference entry.** `allowMachLookup` is absent from the README today (`grep -c allowMachLookup README.md` -> 0), so documenting only the new sibling would be the inconsistent half of that fix. Worth noting the flip side: every other option with a Security Limitations bullet *is* also in the option list, so this is the first that is not. Happy to add entries for both in this PR if you would rather have it that way.
- **No broad `(allow mach*)` browser mode** (cf. #173). The issue asks for the narrow per-service-name option, which is the safer default for a sandbox runtime.
- **No Linux/Windows work** — `mach-register` is Seatbelt-only, and the option is marked macOS-only in its `.describe()`, exactly as `allowMachLookup` is.

One caveat I would rather state than leave implied: I proved the mechanism with a `bootstrap_check_in` probe, not by launching Playwright end-to-end. This removes the blocker the issue documents and is the option it asks for, but I have not verified that Chromium needs nothing further.

Supersedes the withdrawn #407, which duplicated the zod refine verbatim rather than factoring it.

## Overlap with #597

I have another PR open against this repo, #597 (host-allowlist deny reasons reaching the in-band 403). Both touch `src/sandbox/sandbox-manager.ts`, in different places: #597 changes `filterNetworkRequest`'s return type and the SOCKS filter, this one adds a five-line `getAllowMachRegister()` accessor and one field in the `wrapWithSandbox` options object. They do not conflict semantically; expect at most a trivial textual conflict, and I am happy to rebase either onto the other in whichever order suits you.


README.md
```diff
@@ -885,6 +885,7 @@ Users should be aware of potential risks that come from allowing broad domains l
...
+- Mach Service Registration (macOS): The `network.allowMachRegister` option lets the sandboxed process publish Mach service names via `bootstrap_check_in` (`(allow mach-register ...)`), which Chromium-based browsers need in order to register their `MachPortRendezvousServer` and Crashpad ports. This is a wider grant than the `allowMachLookup` counterpart: looking up only lets the sandboxed process reach a service that already exists, whereas registering publishes a name into the login session's bootstrap namespace, where processes **outside** the sandbox resolve it. A sandboxed command holding a broad pattern can therefore claim a name whose real server has not started yet and receive the IPC that server's clients intended for it. A name already held by a running service cannot be taken over, so the exposure is to not-yet-registered names rather than live ones. Scope the patterns to the prefixes the command genuinely registers (e.g. `"com.google.chrome.for.testing.*"`) instead of `"*"`.
```
carlostapiacl: Register alone isn't enough for Chromium, and the example name doesn't match Playwright's bundled build:

```suggestion
- Mach Service Registration (macOS): The `network.allowMachRegister` option lets the sandboxed process publish Mach service names via `bootstrap_check_in` (`(allow mach-register ...)`), which Chromium-based browsers need in order to register their `MachPortRendezvousServer` and Crashpad ports. This is a wider grant than the `allowMachLookup` counterpart: looking up only lets the sandboxed process reach a service that already exists, whereas registering publishes a name into the login session's bootstrap namespace, where processes **outside** the sandbox resolve it. A sandboxed command holding a broad pattern can therefore claim a name whose real server has not started yet and receive the IPC that server's clients intended for it. A name already held by a running service cannot be taken over, so the exposure is to not-yet-registered names rather than live ones. Scope the patterns to the prefixes the command genuinely registers (e.g. `"com.google.chrome.for.testing.*"`) instead of `"*"`. It only works together with `allowMachLookup` for the same names: registering publishes the port, and the browser's child processes still need to look it up. Without the lookup the browser hangs instead of failing. The name depends on the build: Chrome for Testing uses `com.google.chrome.for.testing.*`, Playwright's bundled Chromium uses `org.chromium.Chromium.*`.
```

src/sandbox/sandbox-config.ts
```diff
@@ -767,22 +783,17 @@ export const NetworkConfigSchema = z.object({
...
+    .array(machServiceNameSchema)
     .optional()
     .describe(
       'macOS only: Additional XPC/Mach service names to allow looking up. Supports trailing-wildcard prefix matching (e.g., "2BUA8C4S2C.com.1password.*"). Needed for tools like 1Password CLI, Playwright, or the iOS Simulator that communicate via XPC.',
     ),
+  allowMachRegister: z
+    .array(machServiceNameSchema)
+    .optional()
+    .describe(
+      'macOS only: XPC/Mach service names the sandboxed process may register (bootstrap_check_in), the server-side counterpart of allowMachLookup and a separate Seatbelt operation. Same trailing-wildcard matching. Needed by Chromium-based browsers (Playwright, Puppeteer), which register their MachPortRendezvousServer and Crashpad ports, e.g. "com.google.chrome.for.testing.*".',
```
carlostapiacl: Same two points, in the schema description:

```suggestion
      'macOS only: XPC/Mach service names the sandboxed process may register (bootstrap_check_in), the server-side counterpart of allowMachLookup and a separate Seatbelt operation. Same trailing-wildcard matching. Needed by Chromium-based browsers (Playwright, Puppeteer), which register their MachPortRendezvousServer and Crashpad ports, e.g. "com.google.chrome.for.testing.*", or "org.chromium.Chromium.*" for Playwright's bundled Chromium. List the same names in allowMachLookup.',
```

carlostapiacl (commented):
Tested this branch end to end with Playwright's bundled Chromium on macOS 15.7, through `srt --settings`. Nice work, it does what it says.

| settings | result |
|---|---|
| neither | aborts: `bootstrap_check_in org.chromium.Chromium.MachPortRendezvousServer.<pid>: Permission denied (1100)` |
| `allowMachRegister` only | starts, then hangs: the child processes never get their ports |
| `allowMachLookup` + `allowMachRegister`, same prefix | works: 3 contexts in a row, and 4 browsers in parallel × 5 contexts with no errors |
| the prefixes from the description | aborts: this build registers `org.chromium.Chromium.*` |

Two small doc suggestions inline: that register needs the matching lookup (without it the failure is a hang, not an error), and that the name depends on the browser build. Happy to send them as a PR against your branch if you'd rather.

For context: until this lands we run Chromium with `--single-process`. With 4 workers one browser dies, and even with 1 worker the test after a failed one sometimes crashes with SIGTRAP. The same tests pass outside the sandbox.

v0ropaev:
Thank you — that table is worth more than the PR description was. Both suggestions are in, and the hang deserved more than a doc line, so I also made it say something.

**`allowMachRegister` without the matching lookup now warns.** Your row two is not a misconfiguration the profile can detect after the fact: `mach-register` and `mach-lookup` are separate Seatbelt operations and the generator emits only the one the option names, so a caller who lists names under `allowMachRegister` alone gets a profile where the browser publishes its port and its own children are then denied the lookup — which, as you found, blocks rather than errors. `srt` now warns at profile generation, naming the uncovered entries:

```
[sandbox-runtime] WARNING: network.allowMachRegister lists "org.chromium.Chromium.*" with no
network.allowMachLookup entry covering it. Registering a Mach name does not make it resolvable —
mach-register and mach-lookup are separate operations — so a Chromium-based browser will publish
its MachPortRendezvousServer and then hang waiting for children that cannot look it up. List the
same names in allowMachLookup.
```

Coverage is by matched set, not string equality: a prefix lookup `"org.chromium.*"` covers a registered `org.chromium.Chromium.MachPortRendezvousServer`, while an exact lookup does **not** cover a registered prefix pattern — that narrower case is the one that would otherwise hang on every name but the one leaf.

What I deliberately did **not** do is make `allowMachRegister` imply the lookup. It would fix the hang silently, but `mach-lookup` is the wider of the two grants and widening it from an option that does not mention it is exactly the kind of surprise this README spends a paragraph warning about. If you'd rather it imply, say so and I'll switch it — it's a two-line change.

**Docs.** Both your suggestions applied, with the sentence about the warning folded in. The build-dependence is now spelled out in the README and in the schema description: Chrome for Testing `com.google.chrome.for.testing.*`, Playwright's bundled Chromium `org.chromium.Chromium.*`.

Six new tests in `macos-seatbelt.test.ts` (34 pass in that file). Two mutations to check they bite: never warning fails 4 of them, and reversing the prefix-coverage direction fails 2. A third mutation survived and was right to — my first predicate had a `!name.endsWith('*')` guard on the exact branch that could never fire, since that branch only runs for a lookup with no trailing `*`. Dead condition, removed.

Full suite is 1164 pass / 46 fail on this branch and 1158 / 46 on the same branch with my changes stashed — the 46 are ripgrep and proxy/SigV4 tests failing on this machine for their own reasons, and the delta is exactly my six.

carlostapiacl:
Thanks, that's more than I asked for. Re-tested 43dca4c with Playwright's Chromium through `srt --settings`:

| settings | warning | result |
|---|---|---|
| register only | yes | hangs, as expected |
| register + lookup, same prefix | no | works |
| register + wider lookup (`org.chromium.*`) | no | works |
| register prefix + exact lookup of one name | yes | hangs, correctly flagged |

On the question: keep it explicit. The warning removes the silent hang, and having `allowMachRegister` widen the lookup without saying so would be the same kind of surprise the README warns about.

One more data point on the suite, from macOS 15.7 x86_64 with bun 1.4.2: 1216 pass, 680 skip, 1 fail, so the +6 matches yours. The one failure (`symlink-boundary`, "points to root") fails the same on `main` here. Other processes on this machine were running under `srt`, whose default `TMPDIR` is `/tmp/claude`, so the directory came back between the test's cleanup and its `ln -s`. Unrelated to this PR. Your 46 look machine-specific too: only that one fails here.

v0ropaev:
Thank you — that four-row table is the thing I could not produce, since I have no Playwright Chromium here. The row that matters most to me is the last one: register prefix with an exact lookup of one name is precisely the narrower-than-the-pattern case the coverage check was written for, and it is good to see it flagged against a real browser rather than against my fixture.

Noted on the design question, and I agree for the same reason. Leaving it explicit.

On the suite: your `symlink-boundary` diagnosis is a better explanation than mine. I had written the 46 off as "environmental" without chasing any of them; `/tmp/claude` coming back between the cleanup and the `ln -s` because other processes on the machine run under `srt` is a specific, checkable cause, and it means at least some of my 46 are probably the same shape rather than genuinely mysterious. Nothing for this PR either way — the delta on both our machines is exactly the six new tests.

